PR #47, opened 2026-03-31
feat(component): Add deterministic component fingerprints

dmcilvaney:

### DIFF
--- a/internal/fingerprint/fingerprint.go
+++ b/internal/fingerprint/fingerprint.go
@@ -1,0 +1,175 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package fingerprint
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"sort"
+	"strconv"
+
+	"github.com/microsoft/azure-linux-dev-tools/internal/global/opctx"
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+	"github.com/mitchellh/hashstructure/v2"
+)
+
+// hashstructureTagName is the struct tag name used by hashstructure to determine
+// field inclusion. Fields tagged with `fingerprint:"-"` are excluded.
+const hashstructureTagName = "fingerprint"
+
+// ComponentIdentity holds the computed fingerprint for a single component plus
+// a breakdown of individual input hashes for debugging.
+type ComponentIdentity struct {
+	// Fingerprint is the overall SHA256 hash combining all inputs.
+	Fingerprint string `json:"fingerprint"`
+	// Inputs provides the individual input hashes that were combined.
+	Inputs ComponentInputs `json:"inputs"`
+}
+
+// ComponentInputs contains the individual input hashes that comprise a component's
+// fingerprint.
+type ComponentInputs struct {
+	// ConfigHash is the hash of the resolved component config fields (uint64 from hashstructure).
+	ConfigHash uint64 `json:"configHash"`
+	// SourceIdentity is the opaque identity string for the component's source.
+	// For local specs this is a content hash; for upstream specs this is a commit hash.
+	SourceIdentity string `json:"sourceIdentity,omitempty"`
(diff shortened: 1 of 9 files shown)


internal/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+	// SourceIdentity is the opaque identity string for the component's source.
+	// For local specs this is a content hash; for upstream specs this is a commit hash.
+	SourceIdentity string `json:"sourceIdentity,omitempty"`
+	// OverlayFileHashes maps overlay index (as string) to a combined hash of the
+	// source file's basename and content. Keyed by index rather than path to avoid
+	// checkout-location dependence.
+	OverlayFileHashes map[string]string `json:"overlayFileHashes,omitempty"`
+	// AffectsCommitCount is the number of "Affects: <component>" commits in the project repo.
+	AffectsCommitCount int `json:"affectsCommitCount"`
+	// Distro is the effective distro name.
```
reubeno: Why does this need to be tracked?

internal/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+// and additional context. The fs parameter is used to read overlay source file
+// contents for hashing; spec content identity is provided via [IdentityOptions.SourceIdentity].
+//
+// This function is a deterministic combiner: given the same resolved inputs it always
+// produces the same fingerprint. It does not resolve source identity or count commits —
+// those are expected to be pre-resolved by the caller and passed via opts.
+func ComputeIdentity(
+	fs opctx.FS,
+	component projectconfig.ComponentConfig,
+	distroRef projectconfig.DistroReference,
```
reubeno: Is this the distro that the component is being built for or the distro that the component's spec may have come from? Can we document and clarify?

internal/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+// because some overlay types (e.g., patch-add) derive the destination filename from
+// it when no explicit 'file' field is set.
+func hashOverlayFiles(
+	fs opctx.FS,
+	overlays []projectconfig.ComponentOverlay,
+) (map[string]string, error) {
+	hashes := make(map[string]string)
+
+	for idx, overlay := range overlays {
+		if overlay.Source == "" {
```
reubeno: Since this logic needs to know specific properties of different overlay types -- is there a way for this logic to live closer to the implementation of the overlay types? It otherwise may not occur to someone (AI or human) to keep this in sync.

reubeno (commented):
Apart from the individual comments, my main question... what's your thoughts on how we can ensure that this algorithm stays in sync with the component and overlay definitions as the latter evolve?

internal/fingerprint/fingerprint.go
dmcilvaney: Sure, I'll have a look

internal/fingerprint/fingerprint.go
dmcilvaney: It should be the effective distro for the component, ie if one is set, pick that, otherwise the global default.

dmcilvaney: That said, we can probably drop it to just ReleaseVer, that might be the only bit we really care about.

internal/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+		for key := range inputs.OverlayFileHashes {
+			keys = append(keys, key)
+		}
+
+		sort.Strings(keys)
+
+		for _, key := range keys {
+			writeField(hasher, "overlay:"+key, inputs.OverlayFileHashes[key])
+		}
+	}
```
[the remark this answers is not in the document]
dmcilvaney: I don't think this matters? The fingerprint is an opaque value, so long as its stable and reproducible the exact ordering seems unimportant.

reubeno (approved):

dmcilvaney:
Rebased to address go.mod conflict.

PR #48, opened 2026-03-31
feat(cli): Add component identity command

dmcilvaney:


dmcilvaney:
Will need to be reworked based on new path.

PR #95, opened 2026-04-11
feat(cli): Add user hints to help fix errors

dmcilvaney:
Currently only supports config loading errors, but should be extended to also support missing cli tools etc.


internal/app/azldev/command.go
```diff
@@ -105,16 +104,15 @@ func runFuncInternal(innerFunc CmdWithExtraArgsFuncType, requireConfig bool) cob
...
-				"!!! Unable to find and load valid Azure Linux project configuration.\n\n" +
-					"Please either use the -C option to specify a path to the root directory " +
+			env.AddFixSuggestion(
+				"Please either use the -C option to specify a path to the root directory " +
 					"of your Azure Linux project/repo, or else run this tool from within a directory " +
-					"tree that contains an 'azldev.toml' file at its root.\n\n" +
-					"Most commands will not function correctly without a valid configuration.\n\n" +
-					"------------------------------------------------------------------\n",
+					"tree that contains an 'azldev.toml' file at its root. " +
+					"-- Most commands will not function correctly without a valid configuration.",
```
dmcilvaney:
```suggestion
					"Most commands will not function correctly without a valid configuration.",
```

scenario/__snapshots__/TestAzlDevWithInvalidConfig_stderr_1.snap
```diff
@@ -3,12 +3,11 @@ Parse error: /workdir/azldev.toml
...
-
-------------------------------------------------------------------
-
 ##:##:## ERR Error: a valid project and configuration are required to execute this command
+##:##:## WRN =================================================================
+##:##:## WRN     fix the configuration error: error loading project configuration:
+failed to load project configuration:
+project config file "/workdir/azldev.toml" contains invalid data:
+toml: expected character =
+##:##:## WRN     Please either use the -C option to specify a path to the root directory of your Azure Linux project/repo, or else run this tool from within a directory tree that contains an 'azldev.toml' file at its root. -- Most commands will not function correctly without a valid configuration.
```
dmcilvaney:
```suggestion
##:##:## WRN     Please either use the -C option to specify a path to the root directory of your Azure Linux project/repo, or else run this tool from within a directory tree that contains an 'azldev.toml' file at its root. Most commands will not function correctly without a valid configuration.
```

scenario/__snapshots__/TestSnapshots_no-config_fix_hint_stderr_1.snap
```diff
@@ -0,0 +1,5 @@
+##:##:## INF No Azure Linux project found; some commands will not be available.
+##:##:## ERR Error: a valid project and configuration are required to execute this command
+##:##:## WRN =================================================================
+##:##:## WRN     Please either use the -C option to specify a path to the root directory of your Azure Linux project/repo, or else run this tool from within a directory tree that contains an 'azldev.toml' file at its root. -- Most commands will not function correctly without a valid configuration.
```
dmcilvaney:
```suggestion
##:##:## WRN     Please either use the -C option to specify a path to the root directory of your Azure Linux project/repo, or else run this tool from within a directory tree that contains an 'azldev.toml' file at its root. Most commands will not function correctly without a valid configuration.
```

internal/app/azldev/env.go
```diff
@@ -279,6 +289,49 @@ func (env *Env) OutputDir() string {
...
 
+// AddFixSuggestion records a human-readable hint that will be printed after an
+// error to help the user resolve the issue. Suggestions are printed in FIFO order.
+func (env *Env) AddFixSuggestion(suggestion string) {
+	env.fixSuggestions = append(env.fixSuggestions, suggestion)
+}
+
+// PrintFixSuggestions prints the current fix suggestions, if any.
+func (env *Env) PrintFixSuggestions() {
+	if len(env.fixSuggestions) > 0 {
```
reubeno: Minor nit (not blocking) -- code would be ever so slightly easier to read by inverting the check and bailing immediately if there are no suggestions -- allowing one less horizontal tab.

internal/app/azldev/env.go
```diff
@@ -279,6 +289,49 @@ func (env *Env) OutputDir() string {
...
+
+		padding := "    "
+		paddingSize := len(padding)
+
+		maxMsgLength := 0
+		for _, suggestion := range env.fixSuggestions {
+			if len(suggestion) > maxMsgLength {
+				maxMsgLength = len(suggestion)
+			}
+		}
```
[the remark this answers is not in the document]
dmcilvaney: good enough for now, we probably want a real output later anyways.

internal/app/azldev/env.go
```diff
@@ -279,6 +289,49 @@ func (env *Env) OutputDir() string {
...
+
+		boxWidth := max(0, min(consoleWidth, paddingSize+maxMsgLength+paddingSize))
+		boxEdgeString := strings.Repeat("=", boxWidth)
+
+		slog.Warn(boxEdgeString)
+
+		for _, suggestion := range env.fixSuggestions {
+			slog.Warn(padding + suggestion)
+		}
```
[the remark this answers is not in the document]
dmcilvaney: as above, good enough for v1

internal/app/azldev/env.go
```diff
@@ -279,6 +289,49 @@ func (env *Env) OutputDir() string {
...
+			}
+		}
+
+		boxWidth := max(0, min(consoleWidth, paddingSize+maxMsgLength+paddingSize))
+		boxEdgeString := strings.Repeat("=", boxWidth)
+
+		slog.Warn(boxEdgeString)
+
+		for _, suggestion := range env.fixSuggestions {
+			slog.Warn(padding + suggestion)
```
reubeno: I'm still not convinced it's worth pushing all this through `slog`, but we can always iterate on it later. (I also think it's worth using color to highlight for easier visual scanning. Also can be something we do later.)

reubeno (approved):

internal/app/azldev/env.go
dmcilvaney: fixed

internal/app/azldev/env.go
dmcilvaney: agree, should be a proper UI element, but works for now.

reubeno (approved):

PR #100, opened 2026-04-15
feat: Require components to opt out of auto release calculation

dmcilvaney:
Instead of failing unexpectedly, force all packages to either use auto release, or explicitly opt out of it.

### DIFF
--- a/docs/user/reference/config/components.md
+++ b/docs/user/reference/config/components.md
@@ -9,7 +9,7 @@ A component definition tells azldev where to find the spec file, how to customiz
 | Field | TOML Key | Type | Required | Description |
 |-------|----------|------|----------|-------------|
 | Spec source | `spec` | [SpecSource](#spec-source) | No | Where to find the spec file for this component. Inherited from distro defaults if not specified. |
-| Release calculation | `release-calculation` | string | No | Controls how the Release tag is managed during rendering. `"auto"` (default) = auto-bump; `"manual"` = skip all automatic Release manipulation |
+| Release config | `release` | [ReleaseConfig](#release-configuration) | No | Controls how the Release tag is managed during rendering |
 | Overlays | `overlays` | array of [Overlay](overlays.md) | No | Modifications to apply to the spec and/or source files |
 | Build config | `build` | [BuildConfig](#build-configuration) | No | Build-time options (macros, conditionals, check config) |
 | Source files | `source-files` | array of [SourceFileReference](#source-file-references) | No | Additional source files to download for this component |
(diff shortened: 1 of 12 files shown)


docs/user/reference/config/components.md
```diff
@@ -9,6 +9,7 @@ A component definition tells azldev where to find the spec file, how to customiz
 | Field | TOML Key | Type | Required | Description |
 |-------|----------|------|----------|-------------|
 | Spec source | `spec` | [SpecSource](#spec-source) | No | Where to find the spec file for this component. Inherited from distro defaults if not specified. |
+| Release calculation | `release-calculation` | string | No | Controls how the Release tag is managed during rendering. `"auto"` (default) = auto-bump; `"manual"` = skip all automatic Release manipulation |
```
reubeno: I meant to ask you, @dmcilvaney ...

For cases where we turn release-calculation manual, is there anything you can think of that we could do to decrease the risk that someone forgets to bump it on change?

There's a few handfuls of specs that don't have the integer release inline, but indirect through some other macro defined earlier on and then the Release value is derived from it (or derived from some other macro, further derived from it). An example is `kernel.spec`. For now, we're setting `azl_pkgrelease` under `build.defines` (say, to `"4"`) in the TOML. What do you think of some way to hint the tooling that this define needs to change anytime *anything* else in the component changes? (Or even better yet, allow some way to bump this define?)

This isn't a "this-PR" thing -- the main reason I bring it up is to see if/where you'd place such hints in the TOML schema and whether "release" ought to be made a full section for extensibility, or if you'd more vote for adding peer fields next to this one?

e.g.,:

```toml
[components.foo.release]
bump-policy = "manual"
custom-release-macro = "azl_pkgrelease"
# ... other stuff here ...
```

reubeno (commented):
I've got one question regarding extensibility of the config .. but otherwise, changes look good to me. I'll let you triage the copilot feedback to see what's relevant.

docs/user/reference/config/components.md
dmcilvaney: Agree, will namespace for future proofing

internal/app/azldev/core/sources/release.go
```diff
@@ -138,21 +129,14 @@ func (p *sourcePreparerImpl) tryBumpStaticRelease(
...
 		// The Release tag does not start with an integer (e.g. %{pkg_release})
-		// and the user did not provide an explicit overlay to set it.
+		// and the user did not set release-calculation to "manual".
 		return fmt.Errorf(
 			"component %#q has a non-standard Release tag value %#q that cannot be auto-bumped; "+
-				"add a \"spec-set-tag\" overlay for the Release tag in the component configuration:\n%w",
+				"set 'release-calculation = \"manual\"' in the component configuration "+
+				"and add a \"spec-set-tag\" overlay for the Release tag if needed:\n%w",
 			component.GetName(), releaseValue, err)
 	}
```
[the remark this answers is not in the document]
dmcilvaney: this is as exepcted

reubeno (approved):

PR #103, opened 2026-04-17
feat(image-boot): add ability to boot livecd-style ISO

reubeno:
Adds `--iso` option to `azldev image boot`.

### DIFF
--- a/internal/app/azldev/cmds/image/boot.go
+++ b/internal/app/azldev/cmds/image/boot.go
@@ -27,8 +27,9 @@ const (
 	tempDirPrefixBoot = "azlboot"
 
 	// Default values for VM configuration.
-	defaultSSHPort = 8888
-	defaultCPUs    = 8
+	defaultSSHPort  = 8888
+	defaultCPUs     = 8
+	defaultDiskSize = "10G"
 
 	// Default hostname for cloud-init metadata.
 	defaultHostname = "azurelinux-vm"
@@ -98,19 +99,34 @@ func QEMUDriver(format string) string {
 
 // ImageBootOptions contains options for the boot command.
 type ImageBootOptions struct {
-	Arch                    qemu.Arch
-	AuthorizedPublicKeyPath string
-	UseDiskRW               bool
-	ImageName               string
-	ImagePath               string
-	Format                  ImageFormat
-	SecureBoot              bool
+	// VM configuration (applies to all boot modes).
+	Arch       qemu.Arch
+	SecureBoot bool
+	SSHPort    uint16
+	CPUs       int
+	Memory     string
+
+	// Image source: at least one must be provided (unless '--iso' is used, in which case
+	// an empty disk is created if no image is specified).
+	ImageName string      // Positional arg: look up image in project config to find its path.
+	ImagePath string      // --image-path: explicit path to a disk image (overrides lookup).
+	Format    ImageFormat // --format: explicit format hint for image-name lookup.
+
+	// --iso: bootable ISO (e.g., livecd, installer, rescue media). Attached as a
+	// bootable CD-ROM alongside the disk. Optional; independent of cloud-init.
+	ISOPath  string
+	DiskSize string // --disk-size: size of the empty qcow2 created when '--iso' is used without a disk.
+
+	// Disk behavior.
+	UseDiskRW bool // --rwdisk: persist writes to the source disk image.
+
+	// Cloud-init configuration. If any credential is provided, a seed ISO is generated
+	// and attached. Whether the booted system consumes it depends on cloud-init being
+	// installed and enabled in the guest (same caveat applies to disk and ISO images).
 	TestUserName            string
 	TestUserPassword        string
 	TestUserPasswordFile    string
-	SSHPort                 uint16
-	CPUs                    int
-	Memory                  string
+	AuthorizedPublicKeyPath string
 }
 
 func bootOnAppInit(_ *azldev.App, parentCmd *cobra.Command) {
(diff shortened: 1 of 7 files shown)


internal/app/azldev/cmds/image/boot.go
```diff
@@ -204,84 +249,288 @@ func addBootFlags(cmd *cobra.Command, options *ImageBootOptions) {
...
 
-	if imagePath == "" {
-		if options.ImageName == "" {
-			return errors.New("either IMAGE_NAME argument or --image-path must be specified")
+	passwordBytes, err := fileutils.ReadFile(env.FS(), options.TestUserPasswordFile)
+	if err != nil {
+		return fmt.Errorf("failed to read password file %#q:\n%w", options.TestUserPasswordFile, err)
+	}
+
+	options.TestUserPassword = strings.TrimSpace(string(passwordBytes))
```
dmcilvaney, line 393: Do we care about empty passwords? Or is that just user-error that we let them figure out?

internal/utils/qemu/qemu.go
```diff
@@ -120,8 +126,34 @@ func (r *Runner) Run(ctx context.Context, options RunOptions) error {
...
+		diskBootIndex = 2
+	}
+
+	qemuArgs = append(qemuArgs,
+		"-device", fmt.Sprintf("scsi-hd,drive=hd,bootindex=%d", diskBootIndex),
+	)
+
+	if options.InstallISOPath != "" {
+		qemuArgs = append(qemuArgs,
+			"-drive", fmt.Sprintf("if=none,id=installcd,file=%s,media=cdrom,readonly=on", options.InstallISOPath),
```
dmcilvaney, line 146: maybe not worth it, but `,` in `options.InstallISOPath` can cause some weird qemu parser issues. I *think* you can escape them with `,,` .

internal/utils/qemu/qemu.go
```diff
@@ -239,6 +271,36 @@ func CheckPrerequisites(ctx opctx.Ctx, arch string) error {
...
+		return fmt.Errorf("'qemu-img' prerequisite check failed:\n%w", err)
+	}
+
+	return nil
+}
+
+// CreateEmptyQcow2 creates an empty qcow2 disk image at the given path with the specified size.
+// The size should be a QEMU-compatible size string (e.g., "10G", "512M").
+func (r *Runner) CreateEmptyQcow2(ctx context.Context, path, size string) error {
+	createCmd := exec.CommandContext(ctx, "qemu-img", "create", "-f", "qcow2", path, size)
```
dmcilvaney, line 291: Do we need to do anything to grab stderr here? I think you might need to do `createCmd.Stderr = buffer` to capture it.

internal/app/azldev/cmds/image/boot.go
```diff
@@ -204,84 +249,288 @@ func addBootFlags(cmd *cobra.Command, options *ImageBootOptions) {
...
+	if needEmptyDisk {
+		slog.Warn("'--rwdisk' has no effect with '--iso' alone; the empty disk is ephemeral")
+	} else {
+		slog.Warn("'--rwdisk' enabled: changes will persist to the source disk image")
+	}
+}
+
+// resolveDiskSource returns the source disk image path and format (empty strings if
+// no source disk was requested — i.e., '--iso' is used alone). If IMAGE_NAME is
+// supplied, its presence in project config is validated regardless of whether
```
dmcilvaney, line 429: The command is using `RunFuncWithoutRequiredConfigWithExtraArgs`, might get some odd behaviour here re: validation.

internal/app/azldev/cmds/image/boot.go
```diff
@@ -578,15 +821,19 @@ func buildCloudInitMetadataIso(env *azldev.Env, options *ImageBootOptions, outpu
 
 // buildCloudInitConfig creates the cloud-init configuration for the test user.
 func buildCloudInitConfig(env *azldev.Env, options *ImageBootOptions) (*cloudinit.Config, error) {
+	hasPassword := options.TestUserPassword != ""
```
dmcilvaney, line 857: Might be worth having some dedicated unit tests for this behaviour in boot_internal_test.go.

internal/app/azldev/cmds/image/boot.go
```diff
@@ -144,26 +171,45 @@ Requirements:
...
-			if env.WorkDir() == "" {
-				return nil, errors.New("work dir must be specified in config")
-			}
+  azldev image boot my-image --test-password-file ~/.azl-test-pw --ssh-port 2222 --memory 8G
 
-			if len(args) > 0 {
-				options.ImageName = args[0]
-			}
+  # Boot from an ISO (livecd / installer) onto a new empty 20G disk
+  azldev image boot --iso ~/Downloads/azurelinux.iso --disk-size 20G
```
dmcilvaney, line 177: Do we expect this disk to persist, or is it ephemeral? Currently it looks like its cleaned up along with the temp dir in `runQEMUBoot()`. If thats intentional, might want to add to the tool description.

internal/app/azldev/cmds/image/boot.go
```diff
@@ -126,15 +142,26 @@ func NewImageBootCmd() *cobra.Command {
 		Short: "Boot an Azure Linux image in a QEMU VM",
 		Long: `Boot an Azure Linux image in a QEMU virtual machine.
 
-This command starts a QEMU VM with the specified disk image, setting up a test user
-via cloud-init for access. SSH is forwarded to the host on the specified port (default 8888).
+This command starts a QEMU VM with the specified disk image and/or bootable ISO.
+SSH is forwarded to the host on the specified port (default 8888). If cloud-init
+credentials are provided, a NoCloud seed ISO is generated and attached; the guest
+will consume it only if cloud-init is installed and enabled.
```
dmcilvaney: Will also use cloud-init if test user is set I think, even if they have no password.

internal/app/azldev/cmds/image/boot.go
reubeno: There's currently a warning displayed if the user has specified neither a test password (explicitly or via file) nor an ssh pub key to authorize -- but we don't block. My thought is to go with that for now -- and iterate if we find folks getting stuck. What's your take?

internal/app/azldev/cmds/image/boot.go
reubeno: I've added some verbiage around it. It's ephemeral for now, but we can do something more interesting if enough folks actually use this.

internal/app/azldev/cmds/image/boot.go
reubeno: Good callout. I looked through but I don't see anything concerning since we're still careful not to use config if not present.

internal/utils/qemu/qemu.go
reubeno: Good catch!

internal/utils/qemu/qemu.go
reubeno: Fair point. I'm going to reject inputs with embedded commas -- I don't expect we should really ever legitimately encounter them.

internal/app/azldev/cmds/image/boot.go
dmcilvaney: works for me

dmcilvaney (approved):

PR #106, opened 2026-04-19
refactor(lockfile): split to per-component lock files

dmcilvaney:
Replace single azldev.lock (map[string]ComponentLock) with per-component files under locks/<name>.lock. Each component gets its own lock file with:
- import-commit: permanent fork point (write-once)
- upstream-commit: current resolved commit
- manual-bump: mass-rebuild counter
- input-fingerprint: render input change detection
- resolution-input-hash: offline staleness (reserved for v2)

Update component update command to read/write individual lock files. Drop shared lock mutex, padding logic, and monolithic file handling.

### DIFF
--- a/internal/lockfile/lockfile.go
+++ b/internal/lockfile/lockfile.go
@@ -4,8 +4,7 @@
 // Package lockfile reads and writes per-component lock files under the locks/
 // directory. Each component gets its own <name>.lock TOML file that pins
 // resolved upstream commits and tracks component identity for deterministic
-// builds. Lock files are managed by [azldev component update] and
-// [azldev component bump].
+// builds. Lock files are managed by [azldev component update].
 package lockfile
 
 import (
--- a/internal/lockfile/lockfile.go
+++ b/internal/lockfile/lockfile.go
@@ -1,14 +1,16 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-// Package lockfile reads and writes azldev.lock files, which pin resolved
-// upstream commit hashes for deterministic builds. The lock file is a TOML
-// file at the project root, managed by [azldev component update].
+// Package lockfile reads and writes per-component lock files under the locks/
+// directory. Each component gets its own <name>.lock TOML file that pins
+// resolved upstream commits and tracks component identity for deterministic
+// builds. Lock files are managed by [azldev component update] and
+// [azldev component bump].
 package lockfile
 
 import (
 	"fmt"
-	"strings"
+	"path/filepath"
 
 	"github.com/microsoft/azure-linux-dev-tools/internal/global/opctx"
 	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileperms"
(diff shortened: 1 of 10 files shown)


internal/lockfile/lockfile.go
```diff
@@ -1,149 +1,148 @@
...
 	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileperms"
 	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileutils"
 	toml "github.com/pelletier/go-toml/v2"
 )
 
-// FileName is the lock file name, placed at the project root.
-const FileName = "azldev.lock"
+// LockDir is the directory under the project root where per-component lock
+// files are stored.
+const LockDir = "locks"
```
reubeno, line 22: Can we shift to getting the lock dir tree name from config? I recognize that we'd already set precedent with azldev.lock, but the more we get into new conventional directory structure I think we should follow suit with the other dirs we have configured (e.g., logs, work, rendered specs).

internal/lockfile/lockfile.go
```diff
@@ -1,149 +1,148 @@
...
 type ComponentLock struct {
-	// UpstreamCommit is the resolved full commit hash from the upstream dist-git.
+	// Version is the lock file format version.
+	Version int `toml:"version" comment:"Managed by azldev component update. Do not edit manually."`
+
+	// ImportCommit is the upstream commit hash at the time of initial import
+	// (fork point). Upstream changelog up to this commit is inherited verbatim.
+	// Write-once: set on first import, never changed afterwards.
+	// Empty for local components.
+	ImportCommit string `toml:"import-commit,omitempty"`
```
reubeno, line 39: Remind me if we still need this if we're not squashing changes? Or is the idea that if we at least gather it we don't have to use it -- but we'll have it.

internal/lockfile/store.go
```diff
@@ -0,0 +1,171 @@
...
+	Remove(componentName string) error
+}
+
+// Compile-time check that Store satisfies both interfaces.
+var (
+	_ LockReader = (*Store)(nil)
+	_ LockWriter = (*Store)(nil)
+)
+
+// Store provides cached access to per-component lock files. It wraps the
```
reubeno, line 44: Did you find the caching was already a requirement due to the component count, or is this more future-looking?

internal/lockfile/lockfile.go
dmcilvaney: We will still need it, this tells us when to start interleaving changes in the changelog.

internal/lockfile/lockfile.go
dmcilvaney: Sure, will do

internal/lockfile/store.go
dmcilvaney: Future looking, since we will start pulling information from the locks (ie commit for prep sources etc.). Figured we may as well get the interface setup now. Happy to workshop the interfaces though.

reubeno (approved):

PR #107, opened 2026-04-20
feat: add component-level publish channel configuration

liunan-ms:
Add `ComponentPublishConfig` to `ComponentConfig` for routing RPM, SRPM, and
debuginfo packages to different publish channels and add backwards-compat support for deprecated 'publish.channel' TOML field.

Key changes:
- Add `ComponentPublishConfig` with `rpm-channel`, `srpm-channe`l, `debuginfo-channel` fields
- Add project-level `DefaultComponentConfig`
- Update `ResolveComponentConfig`: distro defaults → project defaults → group
  defaults (sorted) → component explicit config
- Implement `ResolvePackagePublishChannel`: selects channel from resolved component
  config, with package-group and per-package overrides
- Add `IsDebugInfoPackage` helper using hyphen-delimited segment matching
- Expose publish channel in `azldev package list` output
- Remove support for components.<name>.default-package-config (schema + struct field + resolution logic)
- Add `EffectiveRPMChannel()` value method: returns RPMChannel if set, falls back to DeprecatedChannel
- Update docs: project.md (new default-component-config section), components.md,
  component-groups.md, package-groups.md, config-system.md, and how-to guides"

Example output with `components-publish.toml`:
```
# Default everything to SDK
[default-component-config.publish]
rpm-channel = "rpm-sdk"
srpm-channel = "rpm-sdk-srpm"
debuginfo-channel = "rpm-sdk-debuginfo"

[component-groups.base-published]
description = "Components whose packages are published to the rpm-base channel"
components = [
    "cmake",
    "jq"
]
[component-groups.base-published.default-component-config.publish]
srpm-channel = "rpm-base-srpm"
rpm-channel = "rpm-base"
debuginfo-channel = "rpm-base-debuginfo"

# Overrides for specific sub-packages that need to go elsewhere
[package-groups.sdk-published]
packages = [
    "cmake-gui"
]

[package-groups.sdk-published.default-package-config.publish]
rpm-channel = "rpms-sdk"
debuginfo-channel = "rpms-sdk-debuginfo"

```
image
image

### DIFF
--- a/docs/user/reference/config/component-groups.md
+++ b/docs/user/reference/config/component-groups.md
@@ -56,9 +56,10 @@ defines = { azure = "1" }
 
 When a component belongs to one or more groups, the effective configuration is assembled in this order (later layers override earlier ones):
 
-1. Distro version `default-component-config`
-2. Component group `default-component-config` (in alphabetical order by group name if multiple groups apply)
-3. Component's own explicit configuration
+1. Project-level `default-component-config`
+2. Distro version `default-component-config`
+3. Component group `default-component-config` (in alphabetical order by group name if multiple groups apply)
+4. Component's own explicit configuration
 
 See [Configuration Inheritance](../../explanation/config-system.md#configuration-inheritance) for the full details.
 
--- a/docs/user/reference/config/component-groups.md
+++ b/docs/user/reference/config/component-groups.md
@@ -56,8 +56,8 @@ defines = { azure = "1" }
 
 When a component belongs to one or more groups, the effective configuration is assembled in this order (later layers override earlier ones):
 
-1. Project-level `default-component-config`
-2. Distro version `default-component-config`
+1. Distro version `default-component-config`
+2. Project-level `default-component-config`
 3. Component group `default-component-config` (in alphabetical order by group name if multiple groups apply)
 4. Component's own explicit configuration
 
--- a/internal/app/azldev/cmds/pkg/list.go
+++ b/internal/app/azldev/cmds/pkg/list.go
@@ -332,7 +332,7 @@ func synthesizeDebugPackages(
 	// standard package-resolution chain for '<component>-debugsource' so the entry
 	// reflects any explicit package override, package-group defaults, component
 	// settings, or project defaults instead of an implicit empty channel.
-	for compName := range proj.Components {
+	for compName, comp := range proj.Components {
 		if isDebugPackageName(compName) {
 			continue
 		}
@@ -344,14 +344,18 @@ func synthesizeDebugPackages(
 
 		existing[name] = struct{}{}
 
-		// compOf pins the synthesized package to its owning component; groupOf is nil
-		// because -debugsource entries are never members of a package group.
-		result, err := resolvePackageListResult(name, map[string]string{name: compName}, nil, proj)
+		compCopy := comp
+
+		pkgConfig, err := projectconfig.ResolvePackageConfig(name, &compCopy, proj)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to resolve config for synthesized package %#q:\n%w", name, err)
 		}
 
-		results = append(results, result)
+		results = append(results, PackageListResult{
+			PackageName: name,
+			Component:   compName,
+			Channel:     debugChannelName(pkgConfig.Publish.EffectiveRPMChannel()),
+		})
 	}
 
 	return results, nil
--- a/internal/app/azldev/cmds/pkg/list.go
+++ b/internal/app/azldev/cmds/pkg/list.go
@@ -52,10 +52,11 @@ to look up one or more specific packages by exact name — including packages
 that are not explicitly configured (they resolve using only project defaults).
 
 Resolution order (lowest to highest priority):
-  1. Project default-package-config
-  2. Package group default-package-config
-  3. Component default-package-config
-  4. Component packages.<name> override`,
+  1. Project default-component-config publish settings
+  2. Component-group default-component-config publish settings
+  3. Component publish settings
+  4. Package-group default-package-config
+  5. Component packages.<name> override`,
 		Example: `  # List all explicitly-configured packages
   azldev package list -a
 
@@ -174,28 +175,12 @@ func ListPackages(env *azldev.Env, options *ListPackageOptions) ([]PackageListRe
 	results := make([]PackageListResult, 0, len(toResolve))
 
 	for pkgName := range toResolve {
-		// Resolve using the component's config if one is known; otherwise use an empty config
-		// so that only the project default and group layers are applied.
-		compName := compOf[pkgName]
-		compConfig := &projectconfig.ComponentConfig{}
-
-		if compName != "" {
-			if c, ok := proj.Components[compName]; ok {
-				compConfig = &c
-			}
-		}
-
-		pkgConfig, err := projectconfig.ResolvePackageConfig(pkgName, compConfig, proj)
+		result, err := resolvePackageListResult(pkgName, compOf, groupOf, proj)
 		if err != nil {
-			return nil, fmt.Errorf("failed to resolve config for package %#q:\n%w", pkgName, err)
+			return nil, err
 		}
 
-		results = append(results, PackageListResult{
-			PackageName: pkgName,
-			Group:       groupOf[pkgName],
-			Component:   compName,
-			Channel:     pkgConfig.Publish.Channel,
-		})
+		results = append(results, result)
 	}
 
 	if options.SynthesizeDebugPackages {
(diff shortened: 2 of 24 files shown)


internal/projectconfig/project.go
```diff
@@ -26,6 +27,11 @@ type ProjectConfig struct {
 	// Configuration for tools used by azldev.
 	Tools ToolsConfig `toml:"tools,omitempty" json:"tools,omitempty" jsonschema:"title=Tools configuration,description=Configuration for tools used by azldev"`
 
+	// DefaultComponentConfig is the project-wide default applied to every component before any
+	// component-group or component-level config is considered. It is the lowest-priority layer in
+	// the component publish config resolution order.
+	DefaultComponentConfig ComponentConfig `toml:"default-component-config,omitempty" json:"defaultComponentConfig,omitempty" jsonschema:"title=Default component config,description=Project-wide default applied to all components before group and component overrides"`
```
reubeno, line 32: Maybe I missed it, but does this PR consult the `default-component-config` for anything other than publishing info? With it being such a generic construct, we should presumably fold it into any component configuration resolution (not just publishing).

internal/projectconfig/project.go
```diff
@@ -98,6 +108,43 @@ func validatePackageGroupMembership(groups map[string]PackageGroupConfig) error
 	return nil
 }
 
+// validateComponentPublishGroupMembership checks that no component belongs to multiple
+// component groups that define publish settings. If a component is in multiple groups,
+// at most one may have a non-zero [ComponentPublishConfig] in its DefaultComponentConfig
+// to keep publish routing unambiguous.
+func validateComponentPublishGroupMembership(
```
reubeno: Do we think this is still important? It's now very much a special case here that sticks out a bit in the general component config resolution logic. If you think we can live without it -- my preference would be to skip this constraint, and we can always revisit later.

internal/app/azldev/cmds/pkg/list.go
```diff
@@ -173,16 +173,16 @@ func ListPackages(env *azldev.Env, options *ListPackageOptions) ([]PackageListRe
 			}
```
reubeno: Are you planning to also have `azldev component list` now also report publishing information now that we've got info there?

internal/projectconfig/project.go
liunan-ms: You're right! Wired `DefaultComponentConfig` into full component resolution not just publishing info.

internal/projectconfig/project.go
liunan-ms: Yep, it's not important, especially when we only have a default-component-config and a base-published component group with publishing info. I removed this validation.

internal/app/azldev/cmds/pkg/list.go
liunan-ms: `azldev component list` now is able to output the publish channels, e.g.
image

docs/user/reference/config/component-groups.md
```diff
@@ -56,9 +56,10 @@ defines = { azure = "1" }
 
 When a component belongs to one or more groups, the effective configuration is assembled in this order (later layers override earlier ones):
 
-1. Distro version `default-component-config`
-2. Component group `default-component-config` (in alphabetical order by group name if multiple groups apply)
-3. Component's own explicit configuration
+1. Project-level `default-component-config`
```
reubeno: I'm thinking that the distro version configs should not override the project-level ones (i.e., swap 1 and 2). I know we don't have examples of this, but I'd expect that a project would be more specific than the distro defaults. What do you think, @liunan-ms?

If we decide to make that change, I'm okay if you want to do that in a separate fast-follow PR since nothing is depending on this yet.

internal/projectconfig/component.go
```diff
@@ -256,3 +281,96 @@ func (c *ComponentConfig) WithAbsolutePaths(referenceDir string) *ComponentConfi
...
+	// Apply package-group override if this package belongs to one.
+	for _, group := range proj.PackageGroups {
+		if slices.Contains(group.Packages, pkgName) {
+			if groupChannel := packagePublishChannel(&group.DefaultPackageConfig.Publish, isDebugInfo); groupChannel != "" {
+				channel = groupChannel
+			}
+
+			break
+		}
+	}
```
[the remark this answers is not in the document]
liunan-ms: It doesn't impact the performance quite much (validated by building rpms locally), and after switching the new publish config where the package size will be reduced to only a base package list and an exception sdk list.

internal/projectconfig/component.go
```diff
@@ -260,3 +285,96 @@ func (c *ComponentConfig) WithAbsolutePaths(referenceDir string) *ComponentConfi
...
+// names like "kernel-debuginfo-common-x86_64" are correctly identified while
+// unrelated packages like "elfutils-debuginfod" are not.
+func IsDebugInfoPackage(pkgName string) bool {
+	return containsRPMSegment(pkgName, "-debuginfo") || containsRPMSegment(pkgName, "-debugsource")
+}
+
+// containsRPMSegment reports whether pkgName contains segment as a hyphen-delimited
+// segment — i.e. segment is followed by end-of-string or '-'. This prevents
+// "-debuginfo" from matching "-debuginfod".
+func containsRPMSegment(pkgName, segment string) bool {
```
reubeno, line 300: Why are we not just looking at strict suffix of "-debuginfo" or "-debugsource"?

internal/projectconfig/component.go
```diff
@@ -260,3 +285,96 @@ func (c *ComponentConfig) WithAbsolutePaths(referenceDir string) *ComponentConfi
...
+	// packages not covered by any higher-priority source. This matches the old
+	// single-field 'publish.channel' behaviour where the default applied to all packages.
+	channel := packagePublishChannel(&proj.DefaultPackageConfig.Publish, isDebugInfo)
+
+	// Component-level channel overrides the project default.
+	var compChannel string
+	if isDebugInfo {
+		compChannel = comp.Publish.DebugInfoChannel
+	} else {
+		compChannel = comp.Publish.RPMChannel
```
reubeno, line 344: Perhaps I've missed it, but I'm not seeing any way for being able to query the routing for a SRPM for a component. We don't need to block this PR on that but we'll need to figure out a way.

We should chat about this ^^^ but we can work on it in a subsequent PR IMHO.

internal/projectconfig/configfile.go
```diff
@@ -111,6 +116,84 @@ func (f ConfigFile) Validate() error {
...
+// migrateDeprecatedFields promotes deprecated TOML fields to their current equivalents
+// and logs a warning for each one found. This is called immediately after decoding a config
+// file so the rest of the loading pipeline only sees the canonical field names.
+//
+// Currently migrated:
+//   - [PackagePublishConfig.DeprecatedChannel] → [PackagePublishConfig.RPMChannel] in all
+//     contexts (default-package-config, package-group, and per-package).
+//     [PackagePublishConfig.DebugInfoChannel] is left unset; the old TOML format had no
+//     equivalent field for debuginfo packages.
+func migrateDeprecatedFields(cfg *ConfigFile) {
```
reubeno: We may have cases where we'll need to do migration across formats, but in this situation -- since the deprecated field is in the actual config structures, would it be simpler to just update the package publish resolver code to look at the deprecated field as another fallback? It seems like that would just be a few lines.

internal/projectconfig/component.go
reubeno: Ah, I recall now that `azldev component list` should show the SRPM details, right?

We still may need to iterate on the query mechanism, though.

docs/user/reference/config/component-groups.md
liunan-ms: Thanks for pointing this out! I agree with you on distro config should be applied first, will swap 1 and 2.

internal/projectconfig/component.go
liunan-ms: Packages like `kernel-debuginfo-common-x86_64` (we don't build it but Fedora has it) contain `-debuginfo` as an internal segment, not as a suffix. A strict suffix check would miss those and not classify them as debuginfo packages.

internal/projectconfig/configfile.go
liunan-ms: Right, resolving the deprecated channel field within package.go is much simpler. Updated the code.

internal/projectconfig/component.go
```diff
@@ -260,3 +285,97 @@ func (c *ComponentConfig) WithAbsolutePaths(referenceDir string) *ComponentConfi
...
+// packagePublishChannel returns the rpm-channel or debuginfo-channel from publish config
+// depending on whether the package is a debuginfo package. For non-debuginfo packages,
+// it falls back to the deprecated 'channel' field for backwards compatibility.
+func packagePublishChannel(publish *PackagePublishConfig, isDebugInfo bool) string {
+	if isDebugInfo {
+		return publish.DebugInfoChannel
+	}
+
+	return publish.EffectiveRPMChannel()
+}
```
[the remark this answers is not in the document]
liunan-ms: For debuginfo/debugsource packages with the old `publish.channel` field, the publish channel should be obtained by `azldev pkg list` with `--synthesize-debug-packages` flag instead of the `azlde comp build` output.

internal/projectconfig/component.go
reubeno: Nice catch! I didn't know about that.

PR #108, opened 2026-04-21
feat(package list): add --synthesize-debug-packages to package list

reubeno:
Augments `azldev package list` with synthetic `-debuginfo` packages (one per reported package, with `-debuginfo` modifier of the publish channel) and `-debugsource` packages.

This is a **short-term** approach until we're able to properly merge changes that shift us over to an inheritance-driven declaration for where source packages, `debuginfo` packages, etc. should go for a given component (i.e. #107).

### DIFF
--- a/internal/app/azldev/cmds/pkg/list.go
+++ b/internal/app/azldev/cmds/pkg/list.go
@@ -27,9 +27,9 @@ type ListPackageOptions struct {
 	// '-debuginfo' packages (one per reported package, using a parallel publish
 	// channel derived from the original package's publish channel by appending
 	// '-debuginfo', except when the original channel is "" or "none") and synthetic
-	// '-debugsource' packages (one per component in the project configuration, with
-	// no explicit publish channel — so they resolve to the configured default
-	// publishing channel).
+	// '-debugsource' packages (one per component in the project configuration,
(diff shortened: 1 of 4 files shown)


internal/app/azldev/cmds/pkg/list.go
```diff
@@ -186,10 +198,119 @@ func ListPackages(env *azldev.Env, options *ListPackageOptions) ([]PackageListRe
...
+// which packages were requested via '-p'. Components own packages via [ComponentConfig.Packages],
+// but most listed packages come from package groups with no component association — so scoping
+// debugsource emission to the requested package set cannot be done reliably with the current
+// configuration model.
+//
+// Synthetic entries that collide with an already-present (real) package name are skipped so
+// real configuration always wins. Source packages whose names already end in '-debuginfo' or
+// '-debugsource' do not get a doubled suffix synthesized.
+func synthesizeDebugPackages(
+	results []PackageListResult, proj *projectconfig.ProjectConfig,
```
binujp, line 235: thought: The component loop logic and package loop logic could have been in separate functions. The current structure works too particularly since it is not expected to be a long running feature.

binujp (approved):
Looks good!

PR #110, opened 2026-04-22
fix: support upstream specs in component query

### LINKED ISSUE #19: `comp query` fails for upstream-type components with "invalid spec source type: upstream"
### Description

`azldev comp query -p ` fails for any component using the default upstream spec source type (bare inline entries). It incorrectly rejects `upstream` as an invalid spec source type.

### Repro

```bash
# Any bare upstream entry triggers this — not component-specific
azldev comp query -p coreutils
```

### Error

```
ERR Error: failed to parse spec for component "coreutils":
component "coreutils" has invalid spec source type: upstream
```

### Expected

`comp query` should resolve the upstream spec (clone from Fedora dist-git) and display spec details, the same way it works for local-spec components like `azurelinux-release`.

### Workaround

Other commands handle upstream components correctly — `comp list`, `comp prep-sources`, and `comp build` all work fine. Only `comp query` is affected.

### Evidence that upstream resolution works elsewhere

```bash
# All succeed for the same component:
azldev comp list -a | grep coreutils                       # Shows "Upstream: fedora 43"
azldev comp prep-sources -p coreutils -o "$(mktemp -d)"    # Clones and downloads sources
azldev comp build -p coreutils                              # Builds successfully
```

### Likely Cause

`comp query` has a separate code path for resolving spec source types that doesn't handle the implicit/default `upstream` type, while the other subcommands use a shared resolution path that does.

### Expected Changes

comp query should function correctly when querying basic components from https://github.com/microsoft/azurelinux/tree/tomls/base/main

--
_Imported on behalf of @dmcilvaney_

holly-agyei:
Fixes #19

## Summary
- prepare upstream component sources before parsing specs in `component query`
- normalize bare/unspecified spec sources to upstream in this query-only path
- add a regression test covering the default upstream component case

## Test plan
- `GOCACHE=/tmp/azldev-gocache GOMODCACHE=/tmp/azldev-gomodcache go test ./internal/app/azldev/cmds/component -run 'Test(NewComponentQueryCommand|ComponentQueryCmd_NoMatch|QueryComponents_.*)$' -count=1`


holly-agyei:
@reubeno This PR is ready for review. It adds support for upstream/default specs in `component query` by preparing upstream sources before parsing, and it includes regression coverage for the bare upstream component case.

internal/app/azldev/cmds/component/query_test.go
```diff
@@ -80,3 +80,86 @@ func TestQueryComponents_OneComponent(t *testing.T) {
...
+				if len(cmd.Args) >= 5 &&
+					cmd.Args[0] == "git" &&
+					cmd.Args[1] == "-C" &&
+					cmd.Args[3] == "rev-parse" &&
+					cmd.Args[4] == "HEAD" {
+					return "head123abc\n", nil
+				}
+
+				return "name=test-component\nepoch=0\nversion=1.0.0\nrelease=1.azl3\n", nil
+			}
```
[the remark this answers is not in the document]
holly-agyei: @copilot apply changes based on [this feedback](https://github.com/microsoft/azure-linux-dev-tools/pull/110#discussion_r3134359078)

internal/app/azldev/cmds/component/query_test.go
```diff
@@ -80,3 +80,86 @@ func TestQueryComponents_OneComponent(t *testing.T) {
...
+		},
+		{
+			name: "explicit upstream spec source type uses upstream name",
+			component: projectconfig.ComponentConfig{
+				Name: testComponentName,
+				Spec: projectconfig.SpecConfig{
+					SourceType:   projectconfig.SpecSourceTypeUpstream,
+					UpstreamName: testUpstreamName,
+				},
+			},
```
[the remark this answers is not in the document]
holly-agyei: @copilot apply changes based on [this feedback](https://github.com/microsoft/azure-linux-dev-tools/pull/110#discussion_r3134359041)

internal/app/azldev/cmds/component/query.go
```diff
@@ -90,3 +102,54 @@ func QueryComponents(
...
+
+	workDirFactory, err := workdir.NewFactory(env.FS(), env.WorkDir(), env.ConstructionTime())
+	if err != nil {
+		return nil, fmt.Errorf("failed to create work dir factory:\n%w", err)
+	}
+
+	preparedSourcesDir, err := workDirFactory.Create(comp.GetName(), "query-spec")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create work dir for component %#q:\n%w", comp.GetName(), err)
+	}
```
liunan-ms, line 144: [blocking]: The prepared work dir is leaked. `workDirFactory.Create` returns a real directory under` WorkDir()` and doesn't self-clean, and this function never removes `preparedSourcesDir`. Since `query` runs this per component in the loop, every invocation leaves one clone/prepared tree per non-local component behind, growing without bound. `render.go` pairs its staging dir with a deferred cleanup and the same cleanup can be done here:

```
preparedSourcesDir, err := workDirFactory.Create(comp.GetName(), "query-spec")
if err != nil {
    return nil, fmt.Errorf("failed to create work dir for component %#q:\n%w", comp.GetName(), err)
}
defer func() {
    if rmErr := env.FS().RemoveAll(preparedSourcesDir); rmErr != nil {
        slog.Debug("failed to remove query work dir", "path", preparedSourcesDir, "error", rmErr)
    }
}()
```

internal/app/azldev/cmds/component/query.go
```diff
@@ -90,3 +102,54 @@ func QueryComponents(
...
+
+	if err := preparer.PrepareSources(env, componentForPrep, preparedSourcesDir, true /* applyOverlays */); err != nil {
+		return nil, fmt.Errorf("failed to prepare sources for component %#q:\n%w", comp.GetName(), err)
+	}
+
+	preparedConfig := *componentForPrep.GetConfig()
+	preparedConfig.Spec.SourceType = projectconfig.SpecSourceTypeLocal
+	preparedConfig.Spec.Path = filepath.Join(preparedSourcesDir, comp.GetName()+".spec")
```
liunan-ms, line 152: The hardcoded `<name>.spec` join is correct here, but `render.go` already has a `findSpecFile` helper that does this join plus an existence check. Reusing it would keep the two paths consistent and give a clearer error if the spec is missing after preparation.

internal/app/azldev/cmds/component/query_test.go
```diff
@@ -80,3 +80,86 @@ func TestQueryComponents_OneComponent(t *testing.T) {
...
+			options := component.QueryComponentsOptions{
+				ComponentFilter: components.ComponentFilter{
+					ComponentNamePatterns: []string{testComponentName},
+				},
+			}
+
+			results, err := component.QueryComponents(testEnv.Env, &options)
+			require.NoError(t, err)
+			require.Len(t, results, 1)
+
```
liunan-ms, line 160: Given the work-dir leak noted in `query.go`, please add an assertion that the prepared sources dir is cleaned up after `QueryComponents` returns (e.g. assert `WorkDir()` contains no leftover query-spec tree). That locks in the cleanup fix and prevents a future regression. An error-path case (`PrepareSources` failing) would be a nice-to-have too.

liunan-ms (commented):
@holly-agyei  Thanks for contributing to azure-linux-dev-tools! I left some comments to improve the fix.

liunan-ms:
Thanks for the contribution! The issue is resolved, closing this PR.

PR #111, opened 2026-04-22
feat(lock): add structural validation and feature-gated lock checks

dmcilvaney:
Add lock file consistency validation to the component resolver, gated
behind AZLDEV_ENABLE_LOCK_VALIDATION=1 with --skip-lock-validation as
a permanent CLI escape hatch.

Validation runs in FindComponents after resolution:
- Missing/stale lock detection for non-local components
- Orphan lock detection (only on full-project queries, not filtered)
- Actionable fix suggestions (specific 'update <names>' or 'update -a')

Update command enhancements:
- Orphan lock pruning on 'update -a' (guarded against empty sets)
- Cancellation check before saving (Ctrl+C safety)
- Skip lock validation (update is the lock writer)

Also adds:
- Per-component snapshot rejection in config validation (gated)
- Test environment workaround with TODO(lockfiles) for cleanup
- registerGlobalFlags extraction in app.go

### DIFF
--- a/internal/app/azldev/cmds/component/update.go
+++ b/internal/app/azldev/cmds/component/update.go
@@ -41,7 +41,12 @@ distro snapshot time or explicit pin, then records it in locks/<name>.lock.
 Subsequent commands (render, build) use the locked commit for deterministic,
 reproducible results.
 
-Local components are skipped — they have no upstream commit to resolve.`,
+Local components are skipped — they have no upstream commit to resolve.
+
+When updating all components (-a), orphan lock files (locks for components
+that no longer exist in the project config) are automatically pruned.
+Orphan pruning is skipped when updating individual components to avoid
+accidentally removing lock files for components not included in the filter.`,
 		Example: `  # Update all components
   azldev component update -a
 
@@ -52,7 +57,7 @@ Local components are skipped — they have no upstream commit to resolve.`,
   azldev component update -g core`,
 		RunE: azldev.RunFuncWithExtraArgs(func(env *azldev.Env, args []string) (interface{}, error) {
 			// Skip lock validation — update is the lock file writer.
-			env.SetSkipLockValidation(true)
+			options.ComponentFilter.SkipLockValidation = true
 
 			options.ComponentFilter.ComponentNamePatterns = append(
 				args, options.ComponentFilter.ComponentNamePatterns...,
--- a/internal/app/azldev/cmds/component/update.go
+++ b/internal/app/azldev/cmds/component/update.go
@@ -51,6 +51,9 @@ Local components are skipped — they have no upstream commit to resolve.`,
   # Update components in a group
   azldev component update -g core`,
 		RunE: azldev.RunFuncWithExtraArgs(func(env *azldev.Env, args []string) (interface{}, error) {
+			// Skip lock validation — update is the lock file writer.
+			env.SetSkipLockValidation(true)
+
 			options.ComponentFilter.ComponentNamePatterns = append(
 				args, options.ComponentFilter.ComponentNamePatterns...,
 			)
@@ -87,11 +90,11 @@ func UpdateComponents(env *azldev.Env, options *UpdateComponentOptions) ([]Updat
 	}
 
 	comps := resolved.Components()
-	if len(comps) == 0 {
+	if len(comps) == 0 && !options.ComponentFilter.IncludeAllComponents {
 		return nil, errors.New("no components matched the filter")
 	}
 
-	// Resolve upstream commits in parallel.
+	// Resolve upstream commits in parallel (no-op for empty list).
 	store := env.LockStore()
 	if store == nil {
 		return nil, errors.New("no project directory configured; cannot update lock files")
@@ -99,8 +102,13 @@ func UpdateComponents(env *azldev.Env, options *UpdateComponentOptions) ([]Updat
 
 	results := resolveUpstreamCommitsParallel(env, comps, store)
 
-	// Check results and bail on errors/cancellation before saving.
-	if err := checkUpdateResults(env, results); err != nil {
+	// Don't save if the context was cancelled (Ctrl+C).
+	if env.Context().Err() != nil {
+		return results, errors.New("update cancelled; lock files not updated")
+	}
+
+	// Check results and bail on errors before saving.
+	if err := checkUpdateResults(results); err != nil {
 		return results, err
 	}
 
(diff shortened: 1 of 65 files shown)


docs/user/reference/config/components.md
```diff
@@ -98,21 +97,6 @@ spec = { type = "local", path = "azurelinux-release.spec" }
 
 The `path` is relative to the config file that defines the component. Local spec files and any associated source files should be placed alongside the component's `.comp.toml` file.
 
-## Release Configuration
```
dmcilvaney, line 101: bad rebase, fix

internal/app/azldev/core/components/resolver.go
```diff
@@ -463,10 +463,6 @@ func (r *Resolver) createComponentFromConfig(componentConfig *projectconfig.Comp
 			componentConfig.Name, err)
 	}
 
-	if componentConfig.Release.Calculation == "" {
```
dmcilvaney, line 463: bad rebase

internal/app/azldev/core/components/resolver_test.go
```diff
@@ -36,22 +36,14 @@ func setupTestSpec(t *testing.T, testEnv *testutils.TestEnv, path string) projec
 			SourceType: projectconfig.SpecSourceTypeLocal,
 			Path:       path,
 		},
-		Release: projectconfig.ReleaseConfig{
```
dmcilvaney, line 39: bad rebase

internal/app/azldev/core/sources/release.go
```diff
@@ -88,34 +88,42 @@ func BumpStaticRelease(releaseValue string, commitCount int) (string, error) {
 	return fmt.Sprintf("%d%s", newRelease, suffix), nil
 }
 
+// HasUserReleaseOverlay reports whether the given overlay list contains an overlay
```
dmcilvaney: bad rebase

internal/app/azldev/core/sources/release_internal_test.go
```diff
@@ -1,140 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: bad rebase

internal/app/azldev/core/sources/release_test.go
```diff
@@ -115,3 +116,36 @@ func TestGetReleaseTagValue_FileNotFound(t *testing.T) {
 	_, err := sources.GetReleaseTagValue(ctx.FS(), "/nonexistent.spec")
 	require.Error(t, err)
 }
+
+func TestHasUserReleaseOverlay(t *testing.T) {
```
dmcilvaney: bad rebase

internal/app/azldev/env.go
```diff
@@ -94,6 +94,16 @@ type Env struct {
 	// resolve the issue. Printed in FIFO order.
 	fixSuggestions []string
 
+	// skipLockValidation disables lock file validation.
+	//
+	// During rollout, validation is off by default. Enable by setting
```
dmcilvaney: add TODO

internal/projectconfig/component.go
```diff
@@ -108,26 +108,6 @@ func (g ComponentGroupConfig) WithAbsolutePaths(referenceDir string) ComponentGr
 	return result
 }
 
-// ReleaseCalculation controls how the Release tag is managed during rendering.
```
dmcilvaney, line 113: bad rebase

internal/projectconfig/fingerprint_test.go
```diff
@@ -32,7 +32,6 @@ func TestAllFingerprintedFieldsHaveDecision(t *testing.T) {
 		reflect.TypeFor[projectconfig.SpecSource](),
 		reflect.TypeFor[projectconfig.DistroReference](),
 		reflect.TypeFor[projectconfig.SourceFileReference](),
-		reflect.TypeFor[projectconfig.ReleaseConfig](),
```
dmcilvaney, line 35: bad rebase

internal/app/azldev/core/components/resolver.go
```diff
@@ -477,48 +479,44 @@ func (r *Resolver) createComponentFromConfig(componentConfig *projectconfig.Comp
...
+	}
+
+	// Build resolved config map from the component set.
+	resolvedConfigs := make(map[string]projectconfig.ComponentConfig, len(resolved.keys))
+	for _, comp := range resolved.Components() {
+		resolvedConfigs[comp.GetName()] = *comp.GetConfig()
 	}
 
-	return result, nil
+	//nolint:wrapcheck // ValidateAndSuggestFixes already wraps with context.
```
dmcilvaney: avoid nolint, just wrap it.

internal/projectconfig/configfile.go
```diff
@@ -88,6 +89,17 @@ func (f ConfigFile) Validate() error {
...
+	// enabled. Components inherit the snapshot from the distro/group
+	// default-component-config or the project's default-distro. Per-component
+	// snapshots would create non-deterministic builds that the lock file cannot
+	// reliably track. Use an explicit 'upstream-commit' pin instead.
+	//
+	// Gated behind AZLDEV_ENABLE_LOCK_VALIDATION during rollout.
+	//nolint:godox // tracked by TODO(lockfiles) tag.
+	// TODO(lockfiles): remove env var gate once lock validation is stable.
+	lockValidationEnabled := os.Getenv("AZLDEV_ENABLE_LOCK_VALIDATION") == "1"
+
```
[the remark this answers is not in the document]
dmcilvaney: This check is orthogonal to lock validation, we are just using the same envvar to be pragamtic.

internal/app/azldev/cmds/component/update.go
```diff
@@ -109,8 +117,33 @@ func UpdateComponents(env *azldev.Env, options *UpdateComponentOptions) ([]Updat
...
 	}
 
-	// Filter results for table output: only show changed components.
-	return filterChangedResults(results), nil
+	// Prune orphan lock files when updating all components.
+	// Use the resolved component set (not raw config) to include
+	// spec-glob-discovered components that aren't in config directly.
+	// Lock files are version controlled, so pruning is safe even if the
+	// resolved set is empty (e.g., all components removed from config).
+	if options.ComponentFilter.IncludeAllComponents {
```
reubeno, line 130: I'm still a bit uncomfortable with commands whose semantics are different when used with `-a` vs. not. It violates the principle of least surprise.

Can we at minimum indicate it in the command help usage?

internal/app/azldev/core/components/resolver.go
```diff
@@ -43,7 +43,12 @@ func (r *Resolver) FindComponents(filter *ComponentFilter) (components *Componen
 
 	// If we were asked to include all components, then it's not even worth looking at anything else.
 	if filter.IncludeAllComponents {
-		return r.FindAllComponents()
+		allComps, findErr := r.FindAllComponents()
+		if findErr != nil {
+			return allComps, findErr
+		}
+
+		return allComps, r.validateLockFiles(allComps, true)
```
reubeno: Does this mean that `azldev comp list` will fail if the found components have lock issues? Is this one of those cases where we want to figure out how to report the output and also display errors and return non-zero?

internal/app/azldev/core/components/resolver.go
```diff
@@ -493,3 +498,46 @@ func applyInheritedDefaultsToComponent(
...
+
+// validateLockFiles checks lock file consistency against the resolved component
+// set. Skipped when lock validation is disabled on the environment.
+//
+// When checkOrphans is true (i.e., all components are being validated), orphan
+// lock files are also detected. On filtered commands, only missing/stale checks
+// run — orphan detection is a project-wide invariant that would misfire against
+// a subset.
+func (r *Resolver) validateLockFiles(resolved *ComponentSet, checkOrphans bool) error {
+	if r.env.SkipLockValidation() {
```
reubeno: Since `validateLockFiles()` is only called from a function that has a `ComponentFilter`, what do you think about making the skip lock validation option not a tool-global thing but rather a field in ComponentFilter like the others, which is uniformly added to all commands that operate on components. (Filter would be a bit of a misnomer at that point, but the model still fits.)

It would also make it more explicitly threaded through as opposed to a global in the env.

internal/app/azldev/cmds/component/update.go
dmcilvaney: added here and to render

internal/app/azldev/core/components/resolver.go
dmcilvaney: made list skip for now

internal/app/azldev/core/components/resolver.go
dmcilvaney: I've made the change here, but might end up needing to roll it back in a subsequent PR depending on how hard it is to pipe it everywhere it needs to go.

reubeno (approved):

PR #112, opened 2026-04-23
refactor(config): extract ResolveComponentConfig for config inheritance

dmcilvaney:
Extract the distro → group → component config merge logic from applyInheritedDefaultsToComponent into projectconfig.ResolveComponentConfig.

This creates a single source of truth for the inheritance chain that can be reused by lock validation and other consumers without depending on the resolver package.

Pure refactor — no behavior change. Merge order (distro defaults → sorted group defaults → component explicit config) is preserved.


reubeno (approved):
Thank you for doing this!

PR #119, opened 2026-04-24
feat(overlays): make source optional in file-add overlays

### LINKED ISSUE #30: overlays: file-add source field should be optional
The common use case for `file-add` is probably going to be where `file` == `source`, so it would be good for the `source` field to be optional so we don't always have to specify both of them with the same value

--
_Imported on behalf of @ddstreet_

Kelvinoppong:
## Summary
- Makes the `source` field optional for `file-add` overlays.
- When `source` is omitted, it defaults to the value of `file`, covering the common case where destination and source filenames match.
- Removes redundancy in overlay definitions and improves usability.

Closes #30

## Changes
- `internal/projectconfig/overlay.go`: apply the `source`-defaults-to-`file` logic in `WithAbsolutePaths`; relax `Validate()` so `source` is no longer required for `file-add`; update struct tag / JSON schema description.
- `internal/projectconfig/overlay_test.go`: flip the "file-add missing source" case to expect success; add `TestComponentOverlay_WithAbsolutePaths` covering explicit source preserved, default-from-`file`, no-mutation-of-original, and non-`file-add` types untouched.
- `docs/user/reference/config/overlays.md`: document the optional field and add a shorthand example.

## Test plan
- [x] `go test ./internal/projectconfig/...` passes locally
- [x] `go test ./internal/app/azldev/core/sources/...` passes locally
- [x] Generated schema (`schemas/azldev.schema.json`) already reflects the new description (diff verified clean)
- [x] CI scenario / unit / check jobs pass


Kelvinoppong:
@reubeno  this PR is ready for review.

Kelvinoppong:
@reubeno this pr  is ready for review

PR #121, opened 2026-04-24
feat(dist-git): Construct dist-git with lock file history

Tonisal-byte:
Replaces the Affects: commit trailer approach for synthetic dist-git history with lock file fingerprint change detection. Instead of scanning commit messages for `Affects: <component>` lines, synthetic commits are now derived from changes to a component's `input-fingerprint` field in its lock file (specs/<letter>/<name>/<name>.lock).

### DIFF
--- a/internal/app/azldev/cmds/component/component.go
+++ b/internal/app/azldev/cmds/component/component.go
@@ -4,10 +4,7 @@
 package component
 
 import (
-	"path/filepath"
-
 	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev"
-	"github.com/microsoft/azure-linux-dev-tools/internal/lockfile"
 	"github.com/spf13/cobra"
 )
 
@@ -34,17 +31,4 @@ components defined in the project configuration.`,
 	queryOnAppInit(app, cmd)
 	renderOnAppInit(app, cmd)
 	updateOnAppInit(app, cmd)
-}
-
-// relativeLockDir returns the lock file directory relative to the project
-// repository root. Uses the configured 'lock-dir' from the project config
-// if set, falling back to the default [lockfile.LockDir].
-func relativeLockDir(env *azldev.Env) string {
-	if cfg := env.Config(); cfg != nil && cfg.Project.LockDir != "" {
-		if rel, err := filepath.Rel(env.ProjectDir(), cfg.Project.LockDir); err == nil {
-			return rel
-		}
-	}
-
-	return lockfile.LockDir
 }
--- a/internal/app/azldev/cmds/component/component.go
+++ b/internal/app/azldev/cmds/component/component.go
@@ -4,7 +4,10 @@
 package component
 
 import (
+	"path/filepath"
+
 	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev"
+	"github.com/microsoft/azure-linux-dev-tools/internal/lockfile"
 	"github.com/spf13/cobra"
 )
 
@@ -31,4 +34,17 @@ components defined in the project configuration.`,
 	queryOnAppInit(app, cmd)
 	renderOnAppInit(app, cmd)
 	updateOnAppInit(app, cmd)
+}
+
+// relativeLockDir returns the lock file directory relative to the project
+// repository root. Uses the configured 'lock-dir' from the project config
+// if set, falling back to the default [lockfile.LockDir].
+func relativeLockDir(env *azldev.Env) string {
+	if cfg := env.Config(); cfg != nil && cfg.Project.LockDir != "" {
+		if rel, err := filepath.Rel(env.ProjectDir(), cfg.Project.LockDir); err == nil {
+			return rel
+		}
+	}
+
+	return lockfile.LockDir
 }
--- a/internal/app/azldev/cmds/component/render.go
+++ b/internal/app/azldev/cmds/component/render.go
@@ -427,7 +427,7 @@ func prepareComponentSources(
 	// WithSkipLookaside avoids expensive tarball downloads — only spec +
 	// sidecar files are needed for rendering.
 	preparerOpts := []sources.PreparerOption{
-		sources.WithGitRepo(env.Config().Project.DefaultAuthorEmail),
+		sources.WithGitRepo(env),
 		sources.WithSkipLookaside(),
 	}
 
--- a/internal/app/azldev/cmds/component/render.go
+++ b/internal/app/azldev/cmds/component/render.go
@@ -427,7 +427,7 @@ func prepareComponentSources(
 	// WithSkipLookaside avoids expensive tarball downloads — only spec +
 	// sidecar files are needed for rendering.
 	preparerOpts := []sources.PreparerOption{
-		sources.WithGitRepo(env),
+		sources.WithGitRepo(env, relativeLockDir(env)),
 		sources.WithSkipLookaside(),
 	}
 
--- a/internal/app/azldev/cmds/component/render.go
+++ b/internal/app/azldev/cmds/component/render.go
@@ -427,7 +427,7 @@ func prepareComponentSources(
 	// WithSkipLookaside avoids expensive tarball downloads — only spec +
 	// sidecar files are needed for rendering.
 	preparerOpts := []sources.PreparerOption{
-		sources.WithGitRepo(env, relativeLockDir(env)),
+		sources.WithGitRepo(env, env.LockReader()),
 		sources.WithSkipLookaside(),
 	}
 
--- a/internal/app/azldev/core/sources/sourceprep.go
+++ b/internal/app/azldev/core/sources/sourceprep.go
@@ -19,6 +19,7 @@ import (
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev/core/components"
 	"github.com/microsoft/azure-linux-dev-tools/internal/global/opctx"
+	"github.com/microsoft/azure-linux-dev-tools/internal/lockfile"
 	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
 	"github.com/microsoft/azure-linux-dev-tools/internal/providers/sourceproviders"
 	"github.com/microsoft/azure-linux-dev-tools/internal/providers/sourceproviders/fedorasource"
@@ -64,12 +65,16 @@ type PreparerOption func(*sourcePreparerImpl)
 // Without this option, no dist-git is created and synthetic history is skipped.
 //
 // The cmdFactory is used to shell out to git for fingerprint change detection.
+// The lockDir is the lock file directory relative to the project repository
+// root (e.g. "locks").
 func WithGitRepo(
 	cmdFactory opctx.CmdFactory,
+	lockDir string,
 ) PreparerOption {
 	return func(p *sourcePreparerImpl) {
 		p.withGitRepo = true
 		p.cmdFactory = cmdFactory
+		p.lockDir = lockDir
 	}
 }
 
@@ -122,6 +127,11 @@ type sourcePreparerImpl struct {
 	// cmdFactory is used to shell out to git for fingerprint change detection
 	// in the project repository. Set via [WithGitRepo].
 	cmdFactory opctx.CmdFactory
+
+	// lockDir is the lock file directory relative to the project repository
+	// root. Used to locate lock files for fingerprint change detection in
+	// synthetic history generation. Set via [WithGitRepo].
+	lockDir string
 
 	// allowNoHashes, when true, allows source file references without hash
 	// values. Missing hashes are computed from the downloaded files.
@@ -319,7 +329,7 @@ func (p *sourcePreparerImpl) collectOverlays(
 
 // initSourcesRepo initializes a new git repository in sourcesDirPath, stages all files,
 // and creates an initial commit. This is used for components that don't have an upstream
-// dist-git so that Affects commits can still be layered on top.
+// dist-git so that synthetic commits can still be layered on top.
 func initSourcesRepo(sourcesDirPath string) (*gogit.Repository, error) {
 	slog.Info("Initializing git repository for sources", "path", sourcesDirPath)
 
@@ -365,9 +375,9 @@ func (p *sourcePreparerImpl) trySyntheticHistory(
 	config := component.GetConfig()
 	componentName := component.GetName()
 
-	lockRelPath, err := LockFilePath(componentName)
+	lockRelPath, err := lockfile.LockPath(p.lockDir, componentName)
 	if err != nil {
-		return err
+		return fmt.Errorf("resolving lock file path for %#q:\n%w", componentName, err)
 	}
 
 	// Build commit metadata from lock file fingerprint changes.
--- a/internal/app/azldev/core/sources/sourceprep.go
+++ b/internal/app/azldev/core/sources/sourceprep.go
@@ -63,12 +63,13 @@ type PreparerOption func(*sourcePreparerImpl)
 // requires the project configuration to reside inside a git repository.
 // Without this option, no dist-git is created and synthetic history is skipped.
 //
-// The defaultAuthorEmail is used for synthetic changelog entries and commits
-// when no author email is available from git history.
-func WithGitRepo(defaultAuthorEmail string) PreparerOption {
+// The cmdFactory is used to shell out to git for fingerprint change detection.
+func WithGitRepo(
+	cmdFactory opctx.CmdFactory,
+) PreparerOption {
 	return func(p *sourcePreparerImpl) {
 		p.withGitRepo = true
-		p.defaultAuthorEmail = defaultAuthorEmail
+		p.cmdFactory = cmdFactory
 	}
 }
 
@@ -118,9 +119,9 @@ type sourcePreparerImpl struct {
 	// source preparation. Git-tracked files are still fetched.
 	skipLookaside bool
 
-	// defaultAuthorEmail is the email address used for synthetic changelog
-	// entries and commits when no author email is available from git history.
-	defaultAuthorEmail string
+	// cmdFactory is used to shell out to git for fingerprint change detection
+	// in the project repository. Set via [WithGitRepo].
+	cmdFactory opctx.CmdFactory
 
 	// allowNoHashes, when true, allows source file references without hash
 	// values. Missing hashes are computed from the downloaded files.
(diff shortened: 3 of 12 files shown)


internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
+	upstreamCommit  *object.Commit
+	syntheticChange *FingerprintChange
+}
+
+// LockFilePath returns the relative path to a component's lock file within the
+// project repository. The path follows the same letter-prefix convention used by
+// [components.RenderedSpecDir]: specs/<letter>/<name>/<name>.lock.
+// Returns an error if componentName is unsafe (absolute, contains path separators
+// or traversal sequences).
+func LockFilePath(componentName string) (string, error) {
```
dmcilvaney: Is it feasible to use lockfile.LockPath() here instead? Better to have one single source of truth.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
-	commitIter, err := repo.Log(&gogit.LogOptions{From: head.Hash()})
+	prefix := strings.ToLower(componentName[:1])
+
+	return filepath.Join("specs", prefix, componentName, componentName+".lock"), nil
+}
+
+// lockFileFields holds the subset of lock file fields needed for fingerprint
+// change detection. This avoids importing the full [lockfile.ComponentLock]
+// struct and decouples the synthetic history logic from lock file versioning.
+type lockFileFields struct {
```
dmcilvaney: Should switch to the real lockfile.ComponentLock type here.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
 	}
 
-	if len(affectsCommits) == 0 {
-		slog.Info("No commits with Affects marker found; "+
-			"creating default commit",
-			"component", componentName)
+	// If no lock file exists, the component has no overlays — the dist-git
+	// is just the upstream as-is, so no synthetic commits are needed.
+	headFields, lockFileErr := gitShowLockFile(ctx, cmdFactory, projectRepoDir, headHash, lockFileRelPath)
+	if lockFileErr != nil {
```
dmcilvaney: Can you add a todo and envvar check to make this a hard error once we are ready? See the pattern at https://github.com/microsoft/azure-linux-dev-tools/blob/f6ce369f720fba265e2599cb2f491eacf4d7a147/internal/app/azldev/core/components/resolver.go#L37

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
+
+	output, err := git.RunInDir(ctx, cmdFactory, repoDir, "show", ref)
 	if err != nil {
-		return nil, fmt.Errorf("failed to find project repository for config file %#q:\n%w",
-			configFilePath, err)
+		return lockFileFields{}, fmt.Errorf("failed to read lock file at %#q:\n%w", ref, err)
+	}
+
+	var fields lockFileFields
+	if err := toml.Unmarshal([]byte(output), &fields); err != nil {
```
dmcilvaney: Should probably be part of lockfile.Load(), or a similar new function. Pattern can be a .Parse() that takes a stream, and a .Load() that takes a filepath, loads the data, then calls the other Load() func.

internal/utils/git/git.go
```diff
@@ -222,3 +223,30 @@ func WithMetadataOnly() GitOptions {
 		opts.args = append(opts.args, "--no-checkout")
 	}
 }
+
+// RunInDir executes a git command in the given directory and returns its
+// trimmed stdout output. The dir argument is passed via 'git -C dir'.
+func RunInDir(
```
dmcilvaney, line 229: More a high-level question, is there a reason we don't want to use go-git for all our git operations? (to be clear, that would be a todo item for later).

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
-			"component", componentName)
+	// If no lock file exists, the component has no overlays — the dist-git
+	// is just the upstream as-is, so no synthetic commits are needed.
+	headFields, lockFileErr := gitShowLockFile(ctx, cmdFactory, projectRepoDir, headHash, lockFileRelPath)
+	if lockFileErr != nil {
+		slog.Debug("No lock file found at HEAD; skipping synthetic history",
+			"lockFile", lockFileRelPath, "error", lockFileErr)
 
-		commit := defaultOverlayCommit(projectRepo, componentName, defaultAuthorEmail)
+		return nil, "", nil
```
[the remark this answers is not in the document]
dmcilvaney: Agree ^

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
-// unixToTime converts a Unix timestamp to a [time.Time] in UTC.
-func unixToTime(unix int64) time.Time {
-	return time.Unix(unix, 0).UTC()
+// commitMetadataFieldCount is the number of fields expected in the output of
+// 'git log -1 --format=%H%n%an%n%ae%n%at%n%s'.
+const commitMetadataFieldCount = 5
+
+// ParseCommitMetadata parses the output of 'git log -1 --format=%H%n%an%n%ae%n%at%n%s'.
+func ParseCommitMetadata(output string) (CommitMetadata, error) {
+	lines := strings.SplitN(strings.TrimSpace(output), "\n", commitMetadataFieldCount)
```
dmcilvaney: This feels brittle if the message has empty strings etc. Also, I'm not sure there is a guarantee that user names etc. can't have newlines etc. in them. Any reason not to use the `\x00` between each field?

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -391,7 +401,7 @@ func (p *sourcePreparerImpl) trySyntheticHistory(
 
 	if os.IsNotExist(statErr) {
 		slog.Info("No .git directory in sources; initializing repository",
-			"component", component.GetName())
+			"component", componentName)
 
 		if _, err := initSourcesRepo(sourcesDirPath); err != nil {
```
dmcilvaney, line 409: I think this might break since we need to find the upstream commit hashes in the git history or things break. When do we need this codepath? Local packages or something?

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
+
+// --- git CLI helpers ---
+
+// gitLogFileMetadata returns commit metadata (newest-first) for all commits
+// that touched the given file path in the repository at repoDir. Each commit's
+// metadata is separated by a NUL byte in the git log output.
+func gitLogFileMetadata(
+	ctx context.Context, cmdFactory opctx.CmdFactory, repoDir, filePath string,
+) ([]CommitMetadata, error) {
+	output, err := git.RunInDir(ctx, cmdFactory, repoDir,
```
dmcilvaney, line 697: Any reason we can't use go-git for this? Does it not handle individual files?

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
-				AuthorEmail: commit.Author.Email,
-				Timestamp:   commit.Author.When.Unix(),
-				Message:     strings.TrimSpace(commit.Message),
-			})
+	var entries []entry //nolint:prealloc // size not known ahead of time.
+
+	for _, meta := range metas {
+		fields, err := gitShowLockFile(ctx, cmdFactory, projectRepoDir, meta.Hash, lockFileRelPath)
+		if err != nil {
+			slog.Warn("Failed to read lock file at commit; skipping",
```
dmcilvaney: Same as the other place, this should be a hard error, we can make it a warning for now with the todo and envvar.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
-		Message:     "Latest state for " + componentName,
+	if upstreamCommit != "" && !foundUpstream {
+		return nil, fmt.Errorf(
+			"upstream-commit %#q not found in dist-git history; "+
+				"the lock file may reference a commit from a different branch",
+			upstreamCommit)
 	}
+
+	if importCommit != "" && !foundImport {
+		slog.Warn("Import-commit not found in dist-git history; using all collected commits",
```
dmcilvaney: Should probably be a hard error as well

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -30,205 +33,674 @@ type CommitMetadata struct {
...
+			}
+
+			delete(interleavedByUpstream, hash)
+		}
+	}
+
+	// Orphaned changes whose upstream-commit wasn't found are dropped —
+	// they reference an upstream commit outside the known dist-git history.
+	for hash, orphaned := range interleavedByUpstream {
+		slog.Warn("Upstream commit referenced by fingerprint change not found in dist-git history; "+
```
dmcilvaney, line 245: Hard error here as well I think

internal/utils/git/git.go
```diff
@@ -222,3 +223,30 @@ func WithMetadataOnly() GitOptions {
...
+	ctx context.Context, cmdFactory opctx.CmdFactory, dir string, args ...string,
+) (string, error) {
```
dmcilvaney, line 231: I think we would want this to use `GitProviderImpl`.

internal/app/azldev/cmds/component/render.go
```diff
@@ -427,7 +427,7 @@ func prepareComponentSources(
 	// WithSkipLookaside avoids expensive tarball downloads — only spec +
 	// sidecar files are needed for rendering.
 	preparerOpts := []sources.PreparerOption{
-		sources.WithGitRepo(env.Config().Project.DefaultAuthorEmail),
+		sources.WithGitRepo(env),
```
dmcilvaney: Do we need the default email anymore? Should it be removed from the schema?

internal/app/azldev/core/sources/synthistory.go
Tonisal-byte: Yup! go-git doesn't have the capability to handle individual files so it would be a considerable performance hit if we used it instead of shelling out to git directly. I'm working on moving other uses for git cli to go-git given there is no performance overhead.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -118,9 +124,14 @@ type sourcePreparerImpl struct {
...
-	// entries and commits when no author email is available from git history.
-	defaultAuthorEmail string
+	// cmdFactory is used to shell out to git for fingerprint change detection
+	// in the project repository. Set via [WithGitRepo].
+	cmdFactory opctx.CmdFactory
+
+	// lockDir is the lock file directory relative to the project repository
+	// root. Used to locate lock files for fingerprint change detection in
+	// synthetic history generation. Set via [WithGitRepo].
+	lockDir string
```
dmcilvaney: Could this be a `LockReader`, and we add some extra functions to the interface (like get path, etc.).

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -350,37 +361,46 @@ func initSourcesRepo(sourcesDirPath string) (*gogit.Repository, error) {
...
 // Returns a non-nil error if history generation fails.
 func (p *sourcePreparerImpl) trySyntheticHistory(
+	ctx context.Context,
 	component components.Component,
 	sourcesDirPath string,
 ) error {
 	config := component.GetConfig()
+	componentName := component.GetName()
+
+	lockRelPath, err := lockfile.LockPath(p.lockDir, componentName)
```
dmcilvaney: less ciritical, but if we have a store, we could use that to also get this data with a new func. This works as is though.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -31,220 +33,694 @@ type CommitMetadata struct {
...
-// chronologically (oldest first).
-func FindAffectsCommits(repo *gogit.Repository, componentName string) ([]CommitMetadata, error) {
-	// Synthetic history depends on a complete project commit log so Affects
-	// trailers can be discovered reliably.
-	shallowCommits, err := repo.Storer.Shallow()
+// FindFingerprintChanges walks the git log of the project repository for commits
+// that changed the given lock file and returns metadata for each commit where the
+// 'input-fingerprint' field changed. Results are sorted chronologically (oldest
+// first).
+func FindFingerprintChanges(
```
dmcilvaney, line 58: I'll need this for change detection as well, if it makes life easier for you feel free to move it into the lock package. If its not easy to do, I'll do so during my PR instead (important to get this in quickly).

internal/app/azldev/cmds/component/component.go
```diff
@@ -32,3 +35,16 @@ components defined in the project configuration.`,
 	renderOnAppInit(app, cmd)
 	updateOnAppInit(app, cmd)
 }
+
+// relativeLockDir returns the lock file directory relative to the project
+// repository root. Uses the configured 'lock-dir' from the project config
+// if set, falling back to the default [lockfile.LockDir].
+func relativeLockDir(env *azldev.Env) string {
+	if cfg := env.Config(); cfg != nil && cfg.Project.LockDir != "" {
```
dmcilvaney: I don't think project config is guaranteed to be the same as git repo root.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -31,220 +33,694 @@ type CommitMetadata struct {
...
+		return nil, "", fmt.Errorf("failed to get HEAD:\n%w", err)
+	}
+
+	// Read the lock file at HEAD. Distinguish "path does not exist in the
+	// commit tree" (file-not-found) from real failures such as TOML parse
+	// errors or unexpected git object errors — only the former is tolerated.
+	headLock, lockFileErr := showLockFileAtCommit(projectRepo, head.Hash().String(), lockFileRelPath)
+	if lockFileErr != nil {
+		// Real errors (parse failures, git object errors, etc.) are always fatal.
+		if !errors.Is(lockFileErr, object.ErrFileNotFound) {
```
dmcilvaney: Might also get directory not found if no locks have ever been generated for the project before.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -31,220 +33,694 @@ type CommitMetadata struct {
...
+			lockFileRelPath, err)
 	}
 
-	configFilePath := configFile.SourcePath()
-	if configFilePath == "" {
-		return "", fmt.Errorf("component %#q source config file has no path", componentName)
+	if len(fpChanges) == 0 {
+		return nil, "", fmt.Errorf(
+			"lock file %#q exists but has no fingerprint changes; "+
+				"this indicates a corrupt or empty lock file history",
```
dmcilvaney: I think this would also occur during local development (ie haven't committed the lock file yet). MIght need to relax this specific check and give a warning or something instead.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -31,220 +33,694 @@ type CommitMetadata struct {
...
 }
+
+// --- git CLI helpers ---
+
+// gitLogFileMetadata returns commit metadata (newest-first) for all commits
+// that touched the given file path in the repository at repoDir. Fields within
+// each record are separated by NUL (\x00); records are separated by SOH (\x01).
+func gitLogFileMetadata(
+	ctx context.Context, cmdFactory opctx.CmdFactory, repoDir, filePath string,
+) ([]CommitMetadata, error) {
```
dmcilvaney, line 696: Need the same shallow repo detection here to fail if we don't have a history.

internal/app/azldev/cmds/component/component.go
```diff
@@ -32,3 +35,16 @@ components defined in the project configuration.`,
...
+		if rel, err := filepath.Rel(env.ProjectDir(), cfg.Project.LockDir); err == nil {
+			return rel
+		}
+	}
+
+	return lockfile.LockDir
```
dmcilvaney: Fix to the config means this will never happen, shouldn't do defaults here now. 

(https://github.com/microsoft/azure-linux-dev-tools/pull/131)

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -31,220 +33,694 @@ type CommitMetadata struct {
...
+	output, err := git.RunInDir(ctx, cmdFactory, repoDir,
```
dmcilvaney, line 697: Can we add an explanation on why `LogOptions{PathFilter: ...}` doesn't work here (ie too slow for large repos).

internal/app/azldev/cmds/component/component.go
```diff
@@ -32,3 +35,16 @@ components defined in the project configuration.`,
 	renderOnAppInit(app, cmd)
 	updateOnAppInit(app, cmd)
 }
+
+// relativeLockDir returns the lock file directory relative to the project
+// repository root. Uses the configured 'lock-dir' from the project config
+// if set, falling back to the default [lockfile.LockDir].
+func relativeLockDir(env *azldev.Env) string {
```
dmcilvaney: This probably belongs in .env as well, if none of the other changes to the lock architecture remove the need for this funciton.

internal/app/azldev/core/sources/sourceprep.go
Tonisal-byte: Chatted offline, moving this to another PR

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -538,6 +522,83 @@ func buildSyntheticCommits(
...
+	}
+
+	// Tolerate both file-not-found and directory-not-found — the latter
+	// occurs when the locks directory has never been created in the repo.
+	if !errors.Is(lockFileErr, object.ErrFileNotFound) &&
+		!errors.Is(lockFileErr, object.ErrDirectoryNotFound) {
+		return nil, fmt.Errorf("failed to read lock file %#q at HEAD:\n%w",
+			lockFileRelPath, lockFileErr)
+	}
+
```
dmcilvaney, line 585: for future PR: We should gracefully fall back to a placeholder commit ("Uncommited changes") or something liike that. That will allow devs to iterate more easily.

dmcilvaney (approved):

PR #122, opened 2026-04-27
feat: add pytest test suite support

reubeno:
Add config-driven pytest test suites with typed TestSuiteConfig. Test suites are defined in [test-suites] with type = "pytest" and a [pytest] subtable specifying working-dir, test-paths (glob-expanded), and extra-args (with {image-path} placeholder substitution).

Rewrite the 'image test' command to resolve test suites from project config, dispatch to the pytest runner, and support --test-suite filtering, --junit-xml output, and automatic image artifact resolution.

Local LISA-specific test runner code is **temporarily** removed; I have a private branch that will bring it back with the new TOML-driven configuration for local test execution--and plan to follow up with it.

### DIFF
--- a/internal/app/azldev/cmds/image/pytestrunner.go
+++ b/internal/app/azldev/cmds/image/pytestrunner.go
@@ -1,0 +1,356 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package image
+
+import (
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/bmatcuk/doublestar/v4"
+	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev"
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileutils"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/prereqs"
+)
+
+const (
+	// pythonProgram is the Python interpreter used to create venvs and run pytest.
+	pythonProgram = "python3"
+
+	// venvDirName is the name of the venv directory created under the azldev work dir.
+	venvDirName = "pytest-venv"
+
+	// imagePlaceholder is the placeholder token for the image path.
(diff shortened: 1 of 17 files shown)


internal/app/azldev/cmds/image/pytestrunner.go
```diff
@@ -0,0 +1,356 @@
...
+
+	// Substitute placeholders in extra args (never glob-expanded).
+	for _, arg := range pytestConfig.ExtraArgs {
+		args = append(args, replacer.Replace(arg))
+	}
+
+	// Append --junit-xml when requested via CLI.
+	if options.JUnitXMLPath != "" {
+		args = append(args, "--junit-xml", options.JUnitXMLPath)
+	}
```
liunan-ms, line 324: Non-blocker: If I run `azldev image test vm-base --junit-xml results.xml`, I may expect the `results.xml` to be written to `/home/user/azurelinux/results.xml` where I run the test command.
Now pytest resolves results.xml relative to its CWD ( `/home/user/azurelinux/base/images/tests`) and writes to `/home/user/azurelinux/base/images/tests/results.xml`, would it be better to absolutize the JUnitXMLPath before appending?

liunan-ms (approved):
The new pytest-based toml config looks good to me.

internal/app/azldev/cmds/image/pytestrunner.go
reubeno: That's a good call-out! Let me look at that.

internal/app/azldev/cmds/image/test.go
```diff
@@ -43,168 +46,207 @@ func NewImageTestCmd() *cobra.Command {
...
-func testImage(env *azldev.Env, options *ImageTestOptions) error {
-	// Check 1: validate test runner.
-	if err := CheckTestRunner(options.TestRunner); err != nil {
-		return err
+// runImageTest resolves which test suites to run and dispatches each one.
+func runImageTest(env *azldev.Env, options *ImageTestOptions) error {
+	cfg := env.Config()
+	if cfg == nil {
+		return errors.New("no project configuration loaded")
 	}
```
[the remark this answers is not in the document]
reubeno: In practice, this error shouldn't occur -- since this code path won't engage when no config is loaded.

PR #125, opened 2026-04-28
fix(component): support git worktrees in synthetic history generation

### LINKED ISSUE #124: `azldev component render` fails against a git worktree
When running from a git worktree, synthetic history generation fails; we get:

```
16:57:15 ERR Failed to prepare component sources component=less error="preparing sources for `less`:\nfailed to generate synthetic history for component `less`:\nfailed to build synthetic commits:\nfailed to find Affects commits for component `less`:\nfailed to get HEAD reference:\nreference not found"
16:57:15 ERR Some components failed to render errorCount=1 cancelledCount=0
```

---

@Tonisal-byte's suspicion was that we need to be cloning things differently:

> Looking back at the code we attempt to open the project dir like this... I think we would need to  `EnableDotGitCommonDir: true`

```
repo, err := gogit.PlainOpenWithOptions(filepath.Dir(configFilePath), &gogit.PlainOpenOptions{
  DetectDotGit: true,
 })
```

reubeno:
Fixes #124

When running `azldev component render` from inside a git worktree, synthetic history generation failed while resolving HEAD:

```
failed to get HEAD reference:
reference not found
```

A linked worktree's `.git` is a file pointing at `<main>/.git/worktrees/<name>`, and refs/HEAD live in the common git directory. Setting `EnableDotGitCommonDir: true` on the `gogit.PlainOpenOptions` makes go-git resolve the common dir correctly.

The other `gogit.PlainOpen` site in `sourceprep.go` operates on a repository azldev creates itself via `PlainInit`, so it is never a worktree and does not need the same change.

---
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/internal/app/azldev/core/sources/synthistory.go
+++ b/internal/app/azldev/core/sources/synthistory.go
@@ -233,7 +233,8 @@ func resolveConfigFilePath(config *projectconfig.ComponentConfig, componentName 
 // walking up the directory tree.
 func openProjectRepo(configFilePath string) (*gogit.Repository, error) {
 	repo, err := gogit.PlainOpenWithOptions(filepath.Dir(configFilePath), &gogit.PlainOpenOptions{
-		DetectDotGit: true,
+		DetectDotGit:          true,
+		EnableDotGitCommonDir: true,
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Tonisal-byte (approved):
LGTM

PR #129, opened 2026-04-28
feat(lock): populate resolved lock data onto components during resolu…

dmcilvaney:
…tion

Adds a runtime-only Locked *ComponentLockData field to ComponentConfig, populated by the resolver from the lock store at component creation time. This separates user intent (Spec.UpstreamCommit = config pin) from resolved reality (Locked.UpstreamCommit = lock file state).

The Fedora source provider now prefers the locked commit over the config pin in GetComponent and short-circuits ResolveIdentity when a locked commit is present, avoiding unnecessary git clones.

Centralizing lock consumption in the resolver keeps render, build, prepare-sources, and diff-sources oblivious to the lock store — they just read config.Locked.

Key behaviors:
- populateFromLock distinguishes 'not found' (normal) from 'corrupt' (warns) so a malformed lock doesn't silently fall back to live HEAD
- Gated on SpecSourceTypeUpstream so orphaned same-name lock files don't leak into local-component configs
- Resolver.SuppressLockWarnings flag silences the staleness warning during 'component update' (the user is already running the fix)
- Update command clears Locked before re-resolution so the source provider's locked-commit short-circuit doesn't echo back stale values
- EffectiveUpstreamCommit() helper centralizes the locked-beats-pin precedence rule used by GetComponent
- WithAbsolutePaths carries Locked through so resolved configs don't silently lose lock data on subsequent transformations

### DIFF
--- a/internal/app/azldev/core/components/resolver.go
+++ b/internal/app/azldev/core/components/resolver.go
@@ -64,6 +64,8 @@ func (r *Resolver) FindComponents(filter *ComponentFilter) (components *Componen
 			return allComps, findErr
 		}
 
+		r.warnOnLockDrift(allComps)
+
 		return allComps, r.validateLockFiles(allComps, true, skipValidation)
 	}
 
@@ -92,6 +94,8 @@ func (r *Resolver) FindComponents(filter *ComponentFilter) (components *Componen
 			return components, err
 		}
 	}
+
+	r.warnOnLockDrift(components)
 
 	return components, r.validateLockFiles(components, false, skipValidation)
 }
@@ -540,21 +544,6 @@ func (r *Resolver) populateFromLock(config *projectconfig.ComponentConfig) {
 		return
 	}
 
-	// Warn when locked commit differs from explicit config pin - this is a
-	// staleness signal. Validation catches it as an error when enabled, but
-	// during the rollout transition we surface it as a warning. Suppressed
-	// for callers (e.g., 'component update') that are about to refresh the
-	// lock themselves.
-	if !r.SuppressLockWarnings &&
-		config.Spec.UpstreamCommit != "" &&
-		lock.UpstreamCommit != "" &&
-		config.Spec.UpstreamCommit != lock.UpstreamCommit {
-		slog.Warn("Lock differs from config pin - run 'component update' to refresh",
-			"component", config.Name,
-			"configPin", config.Spec.UpstreamCommit,
-			"lockedCommit", lock.UpstreamCommit)
-	}
-
 	config.Locked = &projectconfig.ComponentLockData{
 		UpstreamCommit:   lock.UpstreamCommit,
 		ImportCommit:     lock.ImportCommit,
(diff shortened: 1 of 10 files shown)


internal/providers/sourceproviders/fedorasourceprovider.go
```diff
@@ -108,11 +108,18 @@ func (g *FedoraSourcesProviderImpl) GetComponent(
 
 	gitRepoURL := strings.ReplaceAll(g.distroGitBaseURI, "$pkg", upstreamNameToUse)
 
+	// Use the effective upstream commit (locked > config pin).
+	//nolint:godox // tracked by TODO(lockfiles) tag.
+	// TODO(lockfiles): Once lock validation is default-on, remove the fallback
```
reubeno: Why is this TODO here? I don't see this code looking at UpstreamCommit itself.

internal/projectconfig/component.go
```diff
@@ -149,6 +149,25 @@ type ReleaseConfig struct {
 	Calculation ReleaseCalculation `toml:"calculation,omitempty" json:"calculation,omitempty" validate:"omitempty,oneof=auto manual" jsonschema:"enum=auto,enum=manual,default=auto,title=Release calculation,description=Controls how the Release tag is managed during rendering. Empty or omitted means auto."`
 }
 
+// ComponentLockData holds resolved lock file state attached to a component at
```
reubeno, line 152: `projectconfig/component.go` and its types were not originally intended to hold state derived from sources other than the actual .toml files -- that's why we'd gone through the resolver process that yielded a `Component` from the `ComponentConfig`.

I don't want to block this progress since we need it, but I would like you to file follow-up issues to address this (either by updating the intended design and coming up with a pattern for storing runtime-only state, or by adjusting this code).

internal/app/azldev/core/components/resolver.go
```diff
@@ -479,12 +485,86 @@ func (r *Resolver) createComponentFromConfig(componentConfig *projectconfig.Comp
...
+// locked state automatically via config.Locked.
+//
+// IMPORTANT: This must NEVER overwrite user-specified config values. Lock data
+// goes into the separate Locked field, preserving the manifest/lock boundary:
+// Spec.UpstreamCommit = user intent, Locked.UpstreamCommit = resolved reality.
+func (r *Resolver) populateFromLock(config *projectconfig.ComponentConfig) {
+	// Lock state is only meaningful for upstream components. Local and
+	// unspecified-source components do not have upstream commits to track,
+	// and any orphaned same-name lock file should not leak into their config.
+	if config.Spec.SourceType != projectconfig.SpecSourceTypeUpstream {
```
reubeno, line 515: Will it be odd for non-upstream spec components to have a nil lock data field? I understand some of the lock properties won't apply, but this seems like this might cause surprises (or a lock of nil checks) around.

dmcilvaney: its on my todo list, we will need it for change detection if nothing else

internal/providers/sourceproviders/fedorasourceprovider.go
dmcilvaney: Ah, this is from an earlier version before the function was moved. I'll move the TODO to follow the function.

internal/projectconfig/component.go
dmcilvaney: added to backlog

reubeno (approved):

PR #130, opened 2026-04-28
feat(update): add --bump flag for manual rebuild counter

dmcilvaney:
Add --bump flag to 'component update' that increments ManualBump in the lock file without re-resolving upstream commits. Used for mass-rebuild scenarios (e.g., gcc had a bug, rebuild everything).

The bump increments ManualBump, recomputes the fingerprint, and saves the lock file. This changes the component's identity, triggering a new release without any other input change.

Also adds:
- testing.instructions.md with test conventions (mock components, NoOp wrappers, lock files, external commands, test file naming) based on experience making these changes.

### DIFF
--- a/internal/app/azldev/cmds/component/update.go
+++ b/internal/app/azldev/cmds/component/update.go
@@ -129,7 +129,7 @@ func UpdateComponents(env *azldev.Env, options *UpdateComponentOptions) ([]Updat
 	// existing locks and should not delete entries for components that may
 	// have been removed.
 	if options.Bump {
-		results, bumpErr := bumpComponents(env, store, comps)
+		results, bumpErr := bumpComponents(env, store, comps, options)
 		if bumpErr != nil {
 			return results, bumpErr
 		}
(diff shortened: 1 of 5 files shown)


internal/app/azldev/cmds/component/update.go
```diff
@@ -249,6 +267,97 @@ func saveComponentLocks(env *azldev.Env, store *lockfile.Store, results []Update
...
+			if len(saved) > 0 {
+				slog.Info("Lock files bumped before cancellation", "components", saved)
+			}
+
+			return results, fmt.Errorf("bump cancelled; %d of %d components bumped", len(saved), len(comps))
+		}
+
+		name := comp.GetName()
+
+		// Skip non-upstream components — same as the resolve path.
```
reubeno, line 301: Can we more clearly identify that in the command help -- that this has no effect on non-upstream specs? Does this show up as a warning on stderr as well?

internal/app/azldev/cmds/component/update.go
```diff
@@ -111,6 +124,11 @@ func UpdateComponents(env *azldev.Env, options *UpdateComponentOptions) ([]Updat
 		return nil, errors.New("no project directory configured; cannot update lock files")
 	}
 
+	// --bump: increment ManualBump instead of resolving upstream commits.
```
reubeno: IS the "instead of resolving upstream commits" all over this PR really accurate?

dmcilvaney: should be, I'll double check

internal/app/azldev/cmds/component/update.go
dmcilvaney: I'm going to fix the non-upstream thing soon, should work for all packages then.

reubeno (approved):
I saw you're fixing the non-upstream case; I'm okay if you want to merge this first and then fast follow.

PR #133, opened 2026-04-29
feat(lock): extend lock files to cover local components

dmcilvaney:
Local components (spec-on-disk, no upstream) now get lock files written by 'component update' and read by the resolver, the same as upstream components. Lock files for local components have empty upstream-commit and import-commit fields but populated input-fingerprint, enabling uniform change detection across all component types.

Key changes:
- Rename resolveUpstreamCommitsParallel → resolveSourceIdentitiesParallel and resolveOneUpstreamCommit → resolveOneSourceIdentity to reflect that both source types are now handled
- Drop the upstream-only skip in the resolve loop; local components flow through the same goroutine path using ResolveSourceIdentity (which returns a content hash for local specs)
- Add sourceIdentity field to UpdateResult so saveComponentLocks uses the correct identity (commit hash for upstream, content hash for local) when computing fingerprints
- Drop the upstream-only skip in bumpComponents; local components get their ManualBump incremented with source identity re-resolved from the spec directory
- Remove SpecSourceTypeUpstream gate in populateFromLock so the resolver populates Locked for any component with a lock file
- Add distro resolution guidance to go.instructions.md

New tests:
- Local component writes lock with empty upstream-commit and populated fingerprint
- Local spec content change refreshes fingerprint on re-update
- Bump works for local components (increments ManualBump, recomputes fingerprint)
- Resolver populates Locked for local components with lock files
- Resolver leaves Locked nil for local components without lock files

### DIFF
--- a/internal/app/azldev/cmds/component/update.go
+++ b/internal/app/azldev/cmds/component/update.go
@@ -333,7 +333,7 @@ func bumpComponents(
 		}
 
 		// Determine source identity for fingerprint recomputation.
-		srcIdentity, identityErr := bumpSourceIdentity(env, comp, lock)
+		srcIdentity, identityErr := resolveLockedSourceIdentity(env, comp, lock)
 		if identityErr != nil {
 			return results, identityErr
 		}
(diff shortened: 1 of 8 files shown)


internal/app/azldev/cmds/component/update.go
```diff
@@ -378,6 +377,37 @@ func bumpComponents(
 // checkUpdateErrors returns an error if any component failed to resolve.
```
reubeno: Did this comment get orphaned?

internal/app/azldev/cmds/component/update.go
```diff
@@ -378,6 +377,37 @@ func bumpComponents(
...
 // Does NOT log a summary — call [logUpdateSummary] after saves are complete
 // so that Changed counts include fingerprint-only diffs.
+// bumpSourceIdentity returns the source identity to use when recomputing a
+// bumped component's fingerprint. For upstream components, this is the locked
+// commit (bump doesn't change it). For local components, it re-hashes the
+// spec directory.
+func bumpSourceIdentity(
```
reubeno: We tend to use bump to mean "increment the release" or similar. Is this really bumping or just recomputing?

reubeno (approved):
Two comments that would be great to fix if easy, but change looks good to me otherwise and I'm okay with you not blocking on it if you need to move on.

internal/app/azldev/cmds/component/update.go
dmcilvaney: fixed

internal/app/azldev/cmds/component/update.go
dmcilvaney: fixed

reubeno (approved):

PR #139, opened 2026-04-30
fix(dist-git): fix render staging, import-commit seeding, and commit …

dmcilvaney:
…parsing

- Use project WorkDir for render staging instead of /tmp to avoid filling tmpfs on large batch renders with synthetic git history.
- Seed import-commit on first update so synthetic history walks have a bounded starting point instead of traversing entire repo histories.
- Fix TrimRight -> Trim in commit metadata parsing to handle leading newlines that caused hash parse failures.

### DIFF
--- a/internal/app/azldev/core/sources/synthistory.go
+++ b/internal/app/azldev/core/sources/synthistory.go
@@ -709,7 +709,7 @@ func gitLogFileMetadata(
 	var metas []CommitMetadata //nolint:prealloc // trailing empty block after split.
 
 	for _, block := range blocks {
-		block = strings.TrimRight(block, "\r\n")
+		block = strings.Trim(block, "\r\n")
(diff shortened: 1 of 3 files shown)


reubeno (approved):

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -709,7 +709,7 @@ func gitLogFileMetadata(
 	var metas []CommitMetadata //nolint:prealloc // trailing empty block after split.
 
 	for _, block := range blocks {
-		block = strings.TrimRight(block, "\r\n")
+		block = strings.Trim(block, "\r\n")
```
Tonisal-byte, line 712: Thanks for catching this!

Tonisal-byte (approved):

PR #140, opened 2026-04-30
fix(dist-git): Static release bumping from lock history

Tonisal-byte:
Static release bumping for non-autorelease specs now resets the release counter when the upstream Version changes, matching rpmautospec behavior.

Previously, tryBumpStaticRelease counted all synthetic commits regardless of version changes, producing inflated release numbers. Now it walks fingerprint changes newest-to-oldest, reads the Version tag from the dist-git spec at each upstream commit, and only counts commits since the last version change.

### DIFF
--- a/internal/app/azldev/core/sources/release.go
+++ b/internal/app/azldev/core/sources/release.go
@@ -5,16 +5,25 @@ package sources
 
 import (
 	"fmt"
+	"io"
 	"log/slog"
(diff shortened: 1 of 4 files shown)


dmcilvaney (changes requested):
We will want this, but hold for now: https://dev.azure.com/mariner-org/mariner/_workitems/edit/19671

PR #149, opened 2026-05-03
feat(component): add component changed command

dmcilvaney:
Add 'azldev component changed --from <ref> [--to <ref>]' command that
detects which components changed between two git refs by comparing lock
file fingerprints and rendered sources files.

Features:
- ChangeType enum: added, changed, unchanged, deleted
- SourcesChange detection: true, false, unknown
- Deleted component detection via lock directory walking
- Symmetric direction support (forward and backward comparisons)
- Deterministic sorted output
- Explicit component selection always shows results
- Broad scan (-a) filters unchanged by default
- --include-unchanged flag for full output during broad scans

Includes unit tests (in-memory go-git), command wiring tests, and
scenario tests (real git + azldev binary).


internal/app/azldev/cmds/component/changed.go
```diff
@@ -0,0 +1,522 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading commit %#q:\n%w", commitHash, err)
+	}
+
+	tree, err := commit.Tree()
+	if err != nil {
+		return nil, fmt.Errorf("reading tree for commit %#q:\n%w", commitHash, err)
+	}
+
+	return tree, nil
```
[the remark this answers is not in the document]
dmcilvaney: seems to work fine on tags

internal/app/azldev/cmds/component/changed.go
```diff
@@ -0,0 +1,522 @@
...
+	}
+
+	components.AddComponentFilterOptionsToCommand(cmd, &options.ComponentFilter)
+
+	cmd.Flags().StringVar(&options.From, "from", "", "Git ref to compare from (required)")
+	cmd.Flags().StringVar(&options.To, "to", "HEAD", "Git ref to compare to")
+	cmd.Flags().BoolVar(&options.IncludeUnchanged, "include-unchanged", false,
+		"Include unchanged components in output (only applies to broad -a scans; explicit selections always show status)")
+	cmd.Flags().BoolVar(&options.ForceRecalculate, "force-recalculate", false,
+		"Disable optimizations that skip work for unchanged components")
```
liunan-ms, line 93: Out of curiosity, what would be the use case for `--force-recalculate` in real life?

internal/app/azldev/cmds/component/changed.go
```diff
@@ -0,0 +1,522 @@
...
+			return results[i].Component < results[j].Component
+		})
+
+		return results, nil
+	}
+
+	nonConfigResults, err := buildNonConfigResults(
+		fromLocks, toLocks, configNames, ctx, fromTree, toTree, includeUnchanged,
+		forceRecalculate,
+	)
```
liunan-ms, line 324: buildResults and buildNonConfigResults both implement the similar logic except the components to be iterate dand the changeType override. Would it be better to have a single helper used for both buildResults and buildNonConfigResults scenarios?

internal/app/azldev/cmds/component/changed.go
dmcilvaney: Mostly an escape hatch, its a pattern I'll need in a few tools going forward so I figured I'd add it here. It could also be used to detect someone hand-editing the sources file without changing anything else. That should be caught by the lock and render CI gates for tomls/base/main though.

internal/app/azldev/cmds/component/changed.go
dmcilvaney: They used to be the same func, but it was a parameterized mess, so I split it apart. They do look very similar, but have subtle logic differences that made it tough.

I'll give it another shot, see if I can get them back together, but my gut is its not worth it.

liunan-ms: Thanks Daniel! It's non-blocking. This PR looks good to me.

dmcilvaney: @liunan-ms Does this read better to you? I don't mind it I think.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -540,13 +540,7 @@ func openProjectRepo(
...
 	configFilePath := config.SourceConfigFile.SourcePath()
 
-	repo, err := gogit.PlainOpenWithOptions(
-		filepath.Dir(configFilePath),
-		&gogit.PlainOpenOptions{
-			DetectDotGit:          true,
-			EnableDotGitCommonDir: true,
-		},
-	)
+	repo, err := git.OpenProjectRepo(filepath.Dir(configFilePath))
```
reubeno, line 543: Thank you :smile:

internal/app/azldev/cmds/component/changed.go
liunan-ms: Sorry for the delay! This looks much better. Thanks!

liunan-ms (approved):

PR #150, opened 2026-05-03
feat(lockfile): add ReadAllAtCommit for batch lock file reading

dmcilvaney:
Add ReadAllAtCommit to the lockfile package, which reads and parses all lock files from a directory at a specific git commit. Returns a map of component name to ComponentLock.

- Handles missing lock directory gracefully (empty map, not error)
- Skips non-lock files and dot-prefixed files (matches on-disk loader)
- Returns hard error on unparseable lock files

This enables efficient change detection by doing one tree walk per ref instead of N per-file lookups.

### DIFF
--- a/internal/lockfile/show.go
+++ b/internal/lockfile/show.go
@@ -91,7 +91,7 @@ func cleanLockRelDir(lockRelDir string) (string, error) {
 		return "", fmt.Errorf("lockRelDir must be repo-relative, got absolute path %#q", lockRelDir)
 	}
 
-	if strings.HasPrefix(lockRelDir, "..") {
+	if lockRelDir == ".." || strings.HasPrefix(lockRelDir, "../") {
 		return "", fmt.Errorf("lockRelDir %#q escapes repository root", lockRelDir)
 	}
 
--- a/internal/lockfile/show.go
+++ b/internal/lockfile/show.go
@@ -4,10 +4,14 @@
 package lockfile
 
 import (
+	"errors"
 	"fmt"
+	gopath "path"
+	"strings"
 
 	gogit "github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
 )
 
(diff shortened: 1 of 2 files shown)


internal/lockfile/show.go
```diff
@@ -49,3 +53,122 @@ func ShowAtCommit(
...
+
+// cleanLockRelDir normalizes and validates a repo-relative lock directory path.
+func cleanLockRelDir(lockRelDir string) (string, error) {
+	lockRelDir = gopath.Clean(lockRelDir)
+
+	if gopath.IsAbs(lockRelDir) {
+		return "", fmt.Errorf("lockRelDir must be repo-relative, got absolute path %#q", lockRelDir)
+	}
+
+	if strings.HasPrefix(lockRelDir, "..") {
```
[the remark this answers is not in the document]
reubeno: I do tend to agree.

internal/lockfile/show.go
```diff
@@ -49,3 +53,122 @@ func ShowAtCommit(
...
+				entryPath, commitHash, fileErr)
+		}
+
+		content, contentErr := file.Contents()
+		if contentErr != nil {
+			return nil, fmt.Errorf("failed to read contents of %#q at commit %#q:\n%w",
+				entryPath, commitHash, contentErr)
+		}
+
+		lock, parseErr := Parse([]byte(content))
```
reubeno, line 164: Do we want to fail fast and abort if any single file has errors? What if we are processing a tree that somehow ended up with an unparseable lock file-- is it better to return an error but still yield the lock data that it could extract?

I'm not entirely sure how you'll use this so I'll leave this here and mark it non-blocking.

reubeno (approved):
I left 2 comments (one to +1) a copilot comment but will approve this and let you decide how to handle them.

internal/lockfile/show.go
dmcilvaney: I'd prefer to keep a hard stance here, if this causes problems in practice it's easy to relax the requirement.

PR #155, opened 2026-05-05
feat: add explicit release calculation modes (autorelease, static)

dmcilvaney:
Specs with conditional %autorelease/%else fallbacks confuse the auto- detection logic, causing incorrect Release tag updates. Add two new explicit modes to release.calculation config:

- "autorelease": skip all Release manipulation (rpmautospec handles it)
- "static": always bump the static integer release

Existing "auto" (default) and "manual" modes unchanged.

### DIFF
--- a/internal/app/azldev/core/sources/release.go
+++ b/internal/app/azldev/core/sources/release.go
@@ -120,10 +120,10 @@ func (p *sourcePreparerImpl) tryBumpStaticRelease(
 		return nil
 
 	case projectconfig.ReleaseCalculationStatic:
-		return p.bumpStaticRelease(component, sourcesDirPath, commitCount)
+		return p.readAndBumpRelease(component, sourcesDirPath, commitCount, true)
 
 	case projectconfig.ReleaseCalculationAuto:
-		return p.autoBumpStaticRelease(component, sourcesDirPath, commitCount)
+		return p.readAndBumpRelease(component, sourcesDirPath, commitCount, false)
 
 	default:
 		return fmt.Errorf("component %#q has unknown release calculation mode %#q",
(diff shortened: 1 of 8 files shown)


internal/app/azldev/core/sources/release.go
```diff
@@ -137,10 +156,37 @@ func (p *sourcePreparerImpl) tryBumpStaticRelease(
 		return nil
 	}
 
+	return p.applyStaticBump(component, specPath, releaseValue, commitCount)
+}
+
+// bumpStaticRelease unconditionally bumps the static integer release.
+func (p *sourcePreparerImpl) bumpStaticRelease(
```
reubeno: Do we now have a `bumpStaticRelease` function *and* a `BumpStaticRelease` function? That's going to be very confusing. Can you review back through the functions and make sure the contracts are clear?

internal/projectconfig/component.go
```diff
@@ -133,11 +133,22 @@ func (g ComponentGroupConfig) WithAbsolutePaths(referenceDir string) ComponentGr
...
+	// azldev skips all Release tag manipulation, letting rpmautospec resolve the release
+	// number from git history. Use this for specs with conditional %autorelease/%else
+	// fallbacks that confuse auto-detection.
+	ReleaseCalculationAutorelease ReleaseCalculation = "autorelease"
+
+	// ReleaseCalculationStatic explicitly declares that the spec uses a static integer
+	// release. azldev bumps the integer by the synthetic commit count during rendering.
+	// Use this for specs with conditional Release tags where auto-detection picks the
+	// wrong branch.
+	ReleaseCalculationStatic ReleaseCalculation = "static"
```
reubeno, line 150: Is the expectation that this will cover all static integers (over time) -- even if right now it only covers the case of a single static integer with no dots?

internal/app/azldev/core/sources/release.go
liunan-ms: Non-blocking: bumpStaticRelease and autoBumpStaticRelease are byte-for-byte identical except the autorelease check. A small helper would be good to tighten this.

internal/app/azldev/core/sources/release.go
liunan-ms, line 176: For the case where `calculation = "static"` but `Release: %autorelease` in the spec, this error message seems misleading. In this case, the error message should be `switch to release.calculation = \"autorelease\"` instead of `set release.calculation = \"manual\`?

internal/projectconfig/component.go
dmcilvaney: Yes, this is the "use regex" flow, whatever that might support in the future.

reubeno (approved):

liunan-ms (approved):

PR #165, opened 2026-05-07
feat(pkg): split list 'group' column into 'packageGroups' and 'componentGroups'

liunan-ms:
Currently the `group` field on each `PackageListResult` is the name of the package-group that contains the binary package, or "" if the package isn't in any group. After switching to component-based config, it cannot reflect the component group.

This pull request replaces the single `group` field on PackageListResult with two independent sorted []string fields: `packageGroups` (every package-group whose 'packages' list contains the package; always empty for SRPM rows) and `componentGroups` (every component-group the resolved component belongs to). Both fields are guaranteed non-nil so JSON output is '[]' instead of 'null'.
    
Updates the table output, JSON schema, CLI help, how-to docs, and unit tests accordingly.

Example output of `azldev pkg list --rpm-file <rpm_source_map.json>`:
```json
[
...
 {
    "packageName": "fence-agents",
    "type": "srpm",
    "packageGroups": [],
    "componentGroups": [
      "base-packages"
    ],
    "component": "fence-agents",
    "publishChannel": "rpm-base-srpm"
  },
 {
    "packageName": "fence-agents-amt-ws",
    "type": "rpm",
    "packageGroups": [
      "exceptions-packages"
    ],
    "componentGroups": [
      "base-packages"
    ],
    "component": "fence-agents",
    "publishChannel": "rpm-sdk"
  },
  {
    "packageName": "fence-agents-apc",
    "type": "rpm",
    "packageGroups": [],
    "componentGroups": [
      "base-packages"
    ],
    "component": "fence-agents",
    "publishChannel": "rpm-base"
  },
...
]
```

### DIFF
--- a/docs/user/how-to/inspect-package-config.md
+++ b/docs/user/how-to/inspect-package-config.md
@@ -46,7 +46,7 @@ Example output:
 |--------|----------|
 | **Package** | Binary or source package name (RPM `Name` tag) |
 | **Type** | `rpm` for binary packages, `srpm` for source packages (set when using `--rpm-file`) |
-| **Group** | Package-group whose `packages` list contains this package, if any |
+| **Group** | Package-group whose `packages` list contains this package. Falls back to the resolved component's component-group(s) (sorted, comma-joined) when there is no package-group match, or for SRPM rows. |
 | **Component** | Component that has an explicit `packages.<name>` override for this package, if any |
 | **Publish Channel** | Effective publish channel after all config layers are applied |
 
--- a/docs/user/how-to/inspect-package-config.md
+++ b/docs/user/how-to/inspect-package-config.md
@@ -30,14 +30,14 @@ azldev package list -a
 Example output:
 
 ```
-╭──────────────────┬──────┬────────────────┬───────────┬─────────────────╮
-│ PACKAGE          │ TYPE │ GROUP          │ COMPONENT │ PUBLISH CHANNEL │
-├──────────────────┼──────┼────────────────┼───────────┼─────────────────┤
-│ curl-debugsource │ rpm  │ debug-packages │           │ rpm-debug       │
-│ libcurl          │ rpm  │ base-packages  │           │ rpm-base        │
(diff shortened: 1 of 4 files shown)


dmcilvaney (approved):

Tonisal-byte (approved):

reubeno:
@liunan-ms Is the group actually getting used for anything? Or is it just a diagnostic data point? (I would hope nothing is relying on it for routing correctness.)

liunan-ms:
> @liunan-ms Is the group actually getting used for anything? Or is it just a diagnostic data point? (I would hope nothing is relying on it for routing correctness.)

It's metadata surfaced on `azldev package list` output to help users understand why a package resolved the way it did (which component-group provides the baseline, which package-group overrides it). Nothing replies on it so this change will not break anything.

reubeno:
> > @liunan-ms Is the group actually getting used for anything? Or is it just a diagnostic data point? (I would hope nothing is relying on it for routing correctness.)
> 
> It's metadata surfaced on `azldev package list` output to help users understand why a package resolved the way it did (which component-group provides the baseline, which package-group overrides it). Nothing replies on it so this change will not break anything.

Given that's the case, I'd recommend that we replace 'group' with `groups`, then, and make it a list? We can clarify whether each group is a component or a package.

liunan-ms:
> > > @liunan-ms Is the group actually getting used for anything? Or is it just a diagnostic data point? (I would hope nothing is relying on it for routing correctness.)
> > 
> > 
> > It's metadata surfaced on `azldev package list` output to help users understand why a package resolved the way it did (which component-group provides the baseline, which package-group overrides it). Nothing replies on it so this change will not break anything.
> 
> Given that's the case, I'd recommend that we replace 'group' with `groups`, then, and make it a list? We can clarify whether each group is a component or a package.

Good idea! Instead of having a nested `groups` with two group lists, I just removed 'group' and add 'packageGroups' and 'componentGroups' lists to indicate the group that the package belongs to.

PR #166, opened 2026-05-07
feat: add dirty commit for uncommitted config changes in dist-git

Tonisal-byte:
When preparing sources with a synthetic dist-git, the tool now computes the component's current input fingerprint and compares it against the fingerprint stored in the project repo's HEAD lock file. If they differ, an extra "dirty" synthetic commit is appended so the dist-git reflects the developer's uncommitted config/overlay changes.

This ensures that local edits to component config, overlays, or build settings are represented in the dist-git history without requiring a commit to the project repo first.

### DIFF
--- a/internal/app/azldev/core/sources/sourceprep.go
+++ b/internal/app/azldev/core/sources/sourceprep.go
@@ -18,6 +18,7 @@ import (
 	"github.com/go-git/go-git/v5/plumbing/filemode"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev/core/components"
+	"github.com/microsoft/azure-linux-dev-tools/internal/fingerprint"
 	"github.com/microsoft/azure-linux-dev-tools/internal/global/opctx"
 	"github.com/microsoft/azure-linux-dev-tools/internal/lockfile"
 	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
@@ -66,14 +67,18 @@ type PreparerOption func(*sourcePreparerImpl)
 //
 // The cmdFactory is used to shell out to git for fingerprint change detection.
 // The lockReader provides access to per-component lock files and their directory.
+// The releaseVer is the distro's formal release version (e.g., "4.0"), used to
+// compute the current fingerprint for uncommitted-change detection.
 func WithGitRepo(
 	cmdFactory opctx.CmdFactory,
 	lockReader lockfile.LockReader,
+	releaseVer string,
 ) PreparerOption {
 	return func(p *sourcePreparerImpl) {
 		p.withGitRepo = true
 		p.cmdFactory = cmdFactory
 		p.lockReader = lockReader
+		p.releaseVer = releaseVer
 	}
 }
 
@@ -134,6 +139,10 @@ type sourcePreparerImpl struct {
 	// allowNoHashes, when true, allows source file references without hash
 	// values. Missing hashes are computed from the downloaded files.
 	allowNoHashes bool
+
+	// releaseVer is the distro's formal release version. Used to compute the
+	// current fingerprint for uncommitted-change detection. Set via [WithGitRepo].
+	releaseVer string
 }
 
 // NewPreparer creates a new [SourcePreparer] instance. All positional arguments
@@ -373,9 +382,15 @@ func (p *sourcePreparerImpl) trySyntheticHistory(
 	config := component.GetConfig()
 	componentName := component.GetName()
 
-	// Build commit metadata from lock file fingerprint changes.
+	// Compute the current fingerprint for uncommitted-change detection.
+	currentFingerprint, fpErr := computeCurrentFingerprint(p.fs, config, p.releaseVer)
+	if fpErr != nil {
+		return fmt.Errorf("failed to compute current fingerprint:\n%w", fpErr)
+	}
+
 	changes, importCommit, err := buildSyntheticCommits(
 		ctx, p.cmdFactory, config, componentName, p.lockReader.LockDir(),
+		currentFingerprint,
 	)
 	if err != nil {
 		return fmt.Errorf("failed to build synthetic commits:\n%w", err)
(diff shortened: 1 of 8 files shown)


internal/app/azldev/core/sources/synthistory.go
```diff
@@ -520,9 +535,51 @@ func buildSyntheticCommits(
...
+		"head", headLock.InputFingerprint,
+		"existingChanges", len(existingChanges))
+
+	return &FingerprintChange{
+		CommitMetadata: CommitMetadata{
+			Hash:        "dirty",
+			Author:      "azldev",
+			AuthorEmail: "azldev@local",
+			Timestamp:   time.Now().Unix(),
+			Message:     "Local changes (uncommitted)",
```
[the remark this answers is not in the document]
liunan-ms: have the same question as Copilot, is it okay to have different commit hashes on the same uncommitted state?

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -520,9 +535,51 @@ func buildSyntheticCommits(
...
+func BuildDirtyChange(
+	currentFingerprint string,
+	headLock *lockfile.ComponentLock,
+	existingChanges []FingerprintChange,
+) *FingerprintChange {
+	if currentFingerprint == "" {
+		return nil
+	}
+
+	if currentFingerprint == headLock.InputFingerprint {
```
dmcilvaney, line 561: BuildDirtyChange is exported but dereferences `headLock` without checking nil. Internal callers are safe (guarded at ~line 497), but any external caller or future test will panic.

**Fix:** Add nil guard after the empty-fingerprint check:
`go
if headLock == nil {
    return nil
}
`

Add matching test case:
`go
{
    name:               "nil headLock returns nil",
    currentFingerprint: "sha256:something",
    headLock:           nil,
    wantNil:            true,
},
`

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -520,9 +535,51 @@ func buildSyntheticCommits(
...
+
+	return &FingerprintChange{
+		CommitMetadata: CommitMetadata{
+			Hash:        "dirty",
+			Author:      "azldev",
+			AuthorEmail: "azldev@local",
+			Timestamp:   time.Now().Unix(),
+			Message:     "Local changes (uncommitted)",
+		},
+		UpstreamCommit: headLock.UpstreamCommit,
```
dmcilvaney: Do we want the locked commit here, or the currently available one? ie run `comp update`, then `comp render`, I'd expect to see any commit change in dirty lock file included in the render.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -520,9 +535,51 @@ func buildSyntheticCommits(
...
 	return fpChanges, importCommit, nil
 }
 
+// BuildDirtyChange returns a [FingerprintChange] representing uncommitted
+// config/overlay changes. Returns nil when currentFingerprint is empty or
+// matches the HEAD lock file's fingerprint.
+func BuildDirtyChange(
+	currentFingerprint string,
+	headLock *lockfile.ComponentLock,
+	existingChanges []FingerprintChange,
```
dmcilvaney: existingChanges doesn't seem very useful, only used in a debug print. Since this is a public func, should try to keep the API clean

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -429,6 +441,50 @@ func (p *sourcePreparerImpl) trySyntheticHistory(
 	return nil
 }
 
+// computeCurrentFingerprint computes the current input fingerprint for a
+// component from its resolved config. Returns an empty string for local
+// components or when the source identity cannot be determined — these cases
+// skip uncommitted-change detection gracefully.
+func computeCurrentFingerprint(
```
dmcilvaney, line 485: Should probably add some testing for this func, seems important.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -510,6 +516,15 @@ func buildSyntheticCommits(
 				lockFileRelPath)
 		}
 
+		// Even with no committed fingerprint changes, there may be
+		// uncommitted ("dirty") changes worth representing.
+		if dirty := BuildDirtyChange(currentFingerprint, headLock, fpChanges); dirty != nil {
```
dmcilvaney: This call (and the identical one at line ~540) have identical logic and log messages. Restructure to a single post-walk dirty-detection call:

```go
if len(fpChanges) == 0 && len(shallowCommits) > 0 {
    return nil, "", fmt.Errorf(...)
}

if dirty := BuildDirtyChange(currentFingerprint, headLock); dirty != nil {
    slog.Info("Current fingerprint differs from HEAD lock file; adding dirty entry", ...)
    fpChanges = append(fpChanges, *dirty)
}

if len(fpChanges) == 0 {
    slog.Warn("Lock file has no fingerprint changes; skipping synthetic history", ...)
    return nil, "", nil
}
```

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -66,14 +67,18 @@ type PreparerOption func(*sourcePreparerImpl)
 //
 // The cmdFactory is used to shell out to git for fingerprint change detection.
 // The lockReader provides access to per-component lock files and their directory.
+// The releaseVer is the distro's formal release version (e.g., "4.0"), used to
+// compute the current fingerprint for uncommitted-change detection.
 func WithGitRepo(
 	cmdFactory opctx.CmdFactory,
 	lockReader lockfile.LockReader,
+	releaseVer string,
```
dmcilvaney, line 76: Should we just have withdirtydetection as a separate option rather than forcing all git repo paths to use it? 50/50 on this one.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -66,14 +67,18 @@ type PreparerOption func(*sourcePreparerImpl)
 //
 // The cmdFactory is used to shell out to git for fingerprint change detection.
 // The lockReader provides access to per-component lock files and their directory.
+// The releaseVer is the distro's formal release version (e.g., "4.0"), used to
```
dmcilvaney: Can you make it super clear in the comment this needs to be the resolved distro for the component?

internal/app/azldev/core/sources/sourceprep.go
Tonisal-byte: Since I see this as more of a devex feature I was thinking we might as well have it run on all git repo paths but not hard fail. Of course, we could always just give it a `withoutdirtydetection` flag and let the dev decide if they want it...

internal/app/azldev/core/sources/sourceprep.go
dmcilvaney, line 485: Should this live with the other fingerprint code?

internal/app/azldev/core/sources/sourceprep.go
dmcilvaney: I was thinking less from a dev side (they get what they get), and more from a development side. We can pick which tools support dirty detection programmatically by passing that option or not.

Again, take or leave it.

internal/app/azldev/core/sources/synthistory.go
```diff
@@ -500,29 +506,78 @@ func buildSyntheticCommits(
...
+		"current", currentFingerprint,
+		"head", headLock.InputFingerprint)
+
+	return &FingerprintChange{
+		CommitMetadata: CommitMetadata{
+			Hash:        "dirty",
+			Author:      "azldev",
+			AuthorEmail: "azldev@local",
+			Timestamp:   time.Now().Unix(),
+			Message:     "Local changes (uncommitted)",
```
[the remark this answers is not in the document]
Tonisal-byte: Having a deterministic hash for local changes kind of defeats the purpose of generating the commit in the first place. It's there to indicate something has changed on the HEAD.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -66,14 +67,34 @@ type PreparerOption func(*sourcePreparerImpl)
...
+// fingerprint is compared against the committed lock file; if they differ, a
+// "dirty" synthetic commit is appended to represent the uncommitted changes.
+// Without this option, only committed fingerprint changes produce synthetic commits.
+//
+// This should be enabled for commands that operate on the working tree state
+// (build, render, prepare-sources) and left disabled for commands that should
+// only reflect committed state.
+func WithDirtyDetection() PreparerOption {
+	return func(p *sourcePreparerImpl) {
+		p.dirtyDetection = true
```
dmcilvaney, line 97: Should probably error if git isn't set and we DO set this. NewPreparer is probably the right spot.

dmcilvaney (approved):

PR #167, opened 2026-05-08
fix(cli): Don't return null json when getting empty slices

dmcilvaney:
Removes the need to do:
```
azldev some cmd -O json | jq '(. // []) | .[] | select ...
```

### DIFF
--- a/internal/app/azldev/command.go
+++ b/internal/app/azldev/command.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 
 	"github.com/charmbracelet/x/term"
 	"github.com/mattn/go-isatty"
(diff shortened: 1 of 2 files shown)


internal/app/azldev/command.go
```diff
@@ -233,6 +234,13 @@ func createReflectableOptions(env *Env, format reflectable.Format) *reflectable.
 
 // Displays the results of a command to stdout in JSON format.
 func reportResultsAsJSON(env *Env, results interface{}) error {
+	// Normalize a typed-nil slice to an empty slice so it marshals as `[]`
+	// rather than `null`. This keeps JSON output friendly for downstream
+	// pipelines (e.g., `jq '.[]'` works whether or not there are results).
+	if v := reflect.ValueOf(results); v.IsValid() && v.Kind() == reflect.Slice && v.IsNil() {
```
reubeno: [Non blocking] Is it worth doing the same for maps?

reubeno (approved):
1 non-blocking comment

internal/app/azldev/command.go
dmcilvaney: Sure, I'll do that.

PR #168, opened 2026-05-08
feat(cli): Add check-only option to update

dmcilvaney:


reubeno (approved):
Non-blocking comment: check-only is probably a clearer name in this case, but I can't help but comment that this is exactly what the DryRun interface was added for.

PR #170, opened 2026-05-08
feat(cli): Add check-only option to render

dmcilvaney:


reubeno (approved):
I'm a bit worried about the length and complexity of `render.go`; not to block here, but are there some utility or abstractions that could be removed?

Could we simplify logic by batching up "transactions" of the changes we'd make, and then only apply them in non-check mode (to avoid parallel code paths)? Can we extract out utility helpers that raise the altitude above file/dir operations and more on a "rendered spec cache", which is effectively what base/specs does? Sort of like what you did with the lock store?

PR #171, opened 2026-05-09
feat: allow file replacement in `source-files`

PawelWMS:
### Summary
Adds a `replace-upstream` / `replace-reason` mechanism to the per-component
`[[components.<name>.source-files]]` array, letting a project intentionally
override an entry that already exists in the upstream dist-git `sources`
file (e.g. to ship a locally patched tarball). Today, any filename collision
between a `source-files` entry and an upstream `sources` entry is a hard
error — there is no escape hatch short of renaming the artifact.

### Motivation
Some components need to swap out an upstream tarball for a patched copy
(security fixes, vendor backports, etc.) without renaming the file (the
spec references it by exact name). Silently overwriting would be too easy
to fool, so the new behavior is opt-in, requires an explicit human-readable
reason, validates that the targeted upstream entry actually exists, and
emits an auditable `WARN`-level log on every replacement.

### User-visible changes

New fields on `[[components.<name>.source-files]]`:

| Field              | Type   | Required when                               | Behavior |
|--------------------|--------|---------------------------------------------|----------|
| `replace-upstream` | bool   | optional, defaults to `false`               | When `true`, replaces the same-named upstream entry **in place** (preserving its position in the `sources` file). The matching upstream entry **must exist**, otherwise source preparation fails (guards against stale configs and filename typos). When `false`, a filename collision with an upstream entry remains an error. |
| `replace-reason`   | string | **required** when `replace-upstream = true` | Human-readable explanation. Must be empty when `replace-upstream` is `false`. Captured in the WARN log so every override is auditable. |

Example:

```toml
[[components.example.source-files]]
filename = "example-1.0.tar.gz"
hash = "deadbeefcafe..."
hash-type = "SHA512"
origin = { type = "uri", uri = "https://example.com/example-1.0-patched.tar.gz" }
replace-upstream = true
replace-reason = "patched to fix CVE-2026-0001 before upstream"

### CODE REVIEW

File: internal/app/azldev/core/sources/sourceprep.go [hunk 1]
```diff
@@ -614,46 +602,153 @@ func (p *sourcePreparerImpl) readSourcesFileIfExists(sourcesFilePath string) (st
...
+	}
 
-		hash, hashType, err := p.resolveSourceHash(ref, componentName, outputDir)
-		if err != nil {
-			return nil, err
+	// Second pass: emit upstream entries in their original order, swapping in any
+	// in-place replacements; then append the brand-new entries.
+	mergedLines = make([]string, 0, len(existingEntries)+len(appendLines))
+
+	for _, entry := range existingEntries {
```
dmcilvaney: I think this drops comments and reformats existing lines. Don't care about existing lines, but we might want to keep the comments.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -556,28 +564,12 @@ func (p *sourcePreparerImpl) updateSourcesFile(component components.Component, o
 		return fmt.Errorf("failed to parse existing sources file %#q:\n%w", sourcesFilePath, err)
 	}
 
-	existingFilenames := make(map[string]bool, len(existingEntries))
-	for _, entry := range existingEntries {
-		if existingFilenames[entry.Filename] {
-			return fmt.Errorf(
```
dmcilvaney, line 650: Is changing this error to a warning log intentional (from ReadSourcesFileEntries).

internal/projectconfig/configfile.go
```diff
@@ -197,6 +202,28 @@ func validateSourceFiles(sourceFiles []SourceFileReference, componentName string
 	return nil
 }
 
+// validateReplaceUpstream enforces the pairing rules for the 'replace-upstream' and
+// 'replace-reason' fields on a [SourceFileReference]:
+//   - 'replace-upstream = true' requires a non-empty 'replace-reason'.
+//   - 'replace-reason' may only be set when 'replace-upstream = true'.
+func validateReplaceUpstream(ref SourceFileReference, componentName string) error {
+	if ref.ReplaceUpstream && ref.ReplaceReason == "" {
```
dmcilvaney: should do a strip first, catch "   " as well.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -614,46 +602,153 @@ func (p *sourcePreparerImpl) readSourcesFileIfExists(sourcesFilePath string) (st
...
 
-		entry := fedorasource.FormatSourcesEntry(ref.Filename, hashType, hash)
-		newEntries = append(newEntries, entry)
+		mergedLines = append(mergedLines, fedorasource.FormatSourcesEntry(entry.Filename, entry.HashType, entry.Hash))
+	}
+
+	mergedLines = append(mergedLines, appendLines...)
 
-		slog.Debug("New 'sources' file entry",
+	slog.Info("Updating 'sources' file with extra source file entries",
```
dmcilvaney: I think this fires before we actually write anything, should maybe re-phrase the log, or move it.

```suggestion
	slog.Info("Will update 'sources' file with extra source file entries",
```

internal/app/azldev/core/sources/sourceprep_test.go
```diff
@@ -674,6 +675,98 @@ func TestPrepareSources_UpdatesSourcesFile(t *testing.T) {
...
+					ReplaceUpstream: true,
+					ReplaceReason:   "intended to override (typo demo)",
+				},
+			},
+			existingSourcesContent: "SHA512 (existing.tar.gz) = aabbccdd1122\n",
+			expectError:            true,
+			errorContains: []string{
+				"not-in-upstream.tar.gz",
+				"replace-upstream = true",
+				"no entry with that filename exists",
```
dmcilvaney, line 879: Add test for doing multiple replaces.

reubeno:
@PawelWMS Is there ever a time when we would *not* want to replace upstream? I understand that it's there for strictness -- and that's something we've discussed in the past, but will we in practice just always set it to true if it's the same filename as something that's already in the sources file?

PawelWMS:
> @PawelWMS Is there ever a time when we would _not_ want to replace upstream? I understand that it's there for strictness -- and that's something we've discussed in the past, but will we in practice just always set it to true if it's the same filename as something that's already in the sources file?

I think of it more as a way to protect us from accidentally overwriting the original file when the author actually wanted to introduce a new one, and I want us to document/explain why we want to overwrite the original file. And there's a small bonus, that if you made a typo in the file name when you wanted to update the original file, the tool will catch it and tell you about it.

schemas/azldev.schema.json
```diff
@@ -865,6 +865,16 @@
...
+        "replace-upstream": {
+          "type": "boolean",
+          "title": "Replace upstream",
+          "description": "When true intentionally replaces a same-named entry in the upstream 'sources' file. Requires 'replace-reason'."
+        },
+        "replace-reason": {
+          "type": "string",
+          "title": "Replace reason",
+          "description": "Required when 'replace-upstream' is true. Human-readable explanation for the replacement."
         }
```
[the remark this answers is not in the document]
PawelWMS: That's auto-generated by `azldev`, not modified manually. If the auto-generation can be improved to handle optional fields, that's out of scope for this PR.

internal/app/azldev/core/sources/sourceprep_test.go
```diff
@@ -687,6 +694,189 @@ func TestPrepareSources_UpdatesSourcesFile(t *testing.T) {
 			errorContains: []string{
 				"existing.tar.gz",
 				"conflicts with an existing entry",
+				"replace-upstream = true",
```
dmcilvaney, line 697: Might want a test for the combination of `replace-upstream` and `allowNoHashes`, ensure that works as expected.

dmcilvaney (approved):

internal/providers/sourceproviders/fedorasource/fedorasource.go
```diff
@@ -335,58 +354,101 @@ func parseSourcesFile(content string, packageName string, lookasideBaseURI strin
...
+		if previousLine, exists := seen[entry.Filename]; exists {
+			return nil, fmt.Errorf(
+				"duplicate filename %#q in 'sources' file at line %d (already defined at line %d)",
+				entry.Filename, lineNum+1, previousLine)
+		}
 
-			hash = legacyMatches[sourcesLegacyPatternHashValueIndex]
-			fileName = legacyMatches[sourcesLegacyPatternFilenameIndex]
+		seen[entry.Filename] = lineNum + 1
+		parsedLines = append(parsedLines, SourcesFileLine{Raw: rawLine, Entry: &entry})
```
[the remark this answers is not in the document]
PawelWMS: False positive — `entry` is declared with `:=` inside the loop body block (not the `for` header), so each iteration creates a fresh variable; the `&entry` taken on different iterations are distinct heap-escaped pointers. (And since Go 1.22 even the header-scoped form gets per-iteration variables.) The existing `ReadSourcesFile` test verifies this end-to-end by asserting `Entry.Filename` independently for two parsed entries.

internal/app/azldev/core/sources/sourceprep_test.go
PawelWMS: Good idea.

internal/projectconfig/configfile.go
```diff
@@ -197,6 +203,31 @@ func validateSourceFiles(sourceFiles []SourceFileReference, componentName string
...
+				"a non-empty 'replace-reason' is required to document the override",
+			ref.Filename, componentName)
+	}
+
+	if !ref.ReplaceUpstream && trimmedReason != "" {
+		return fmt.Errorf(
+			"source file %#q in component %#q has 'replace-reason' set but 'replace-upstream' is not true; "+
+				"'replace-reason' is only valid when 'replace-upstream = true'",
+			ref.Filename, componentName)
+	}
```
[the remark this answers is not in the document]
PawelWMS: Good catch — fixed in the next push. The `!ReplaceUpstream` branch now checks the raw `ref.ReplaceReason` (not the trimmed value), so any non-empty reason (including whitespace-only) is rejected loudly with the existing "reason set but replace-upstream is not true" error. Added a regression test `TestProjectConfigFileValidation_ReplaceReasonWhitespaceOnlyWithoutReplaceUpstreamRejected` to lock in the behavior.

internal/app/azldev/core/sources/sourceprep_test.go
PawelWMS: Added — see new subtest "replace-upstream with allow-no-hashes computes fresh hash and replaces upstream entry" inside `TestPrepareSources_AllowNoHashes` in `internal/app/azldev/core/sources/sourceprep_test.go`. It seeds the existing `sources` file with a stub upstream hash for the same filename, then verifies the merged file (a) contains the freshly-computed SHA-256 of the on-disk content, (b) preserves the unrelated sibling entry, and (c) no longer contains the original upstream stub hash (proving replacement, not duplication).

PR #173, opened 2026-05-10
refactor(render): migrate phases 1 and 3 to parmap.Map

dmcilvaney:
Replaces the hand-rolled semaphore + waitgroup + resultsChan
pattern in both `parallelPrepare` (phase 1) and
`parallelFinish` (phase 3) with calls to `parmap.Map`.

- Drops the `prepWithSemaphore` indirection in phase 1; the
  worker is now just `prepareOneComponent` (validate + prepare).
- Drops the semaphore parameter from `finishOneComponent` and
  the trailing manual fanout loop in phase 3.
- Removes the `prepResult.index` field — parmap returns results
  in input order, so callers index into them directly.
- Cancellation-before-start (worker never invoked) surfaces via
  `Result.Cancelled` and is mapped to `renderStatusCancelled`
  at the loop callsites, matching the pre-refactor behaviour.
- Drops the unused `sync` import; adds `context` and `parmap`.

No behaviour change. `mage unit` + `mage scenario` pass.

### DIFF
--- a/internal/app/azldev/cmds/component/render.go
+++ b/internal/app/azldev/cmds/component/render.go
@@ -4,6 +4,7 @@
 package component
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


dmcilvaney:
@copilot resolve the merge conflicts in this pull request

reubeno (approved):

PR #181, opened 2026-05-12
feat(render): add --allow-dirty flag for uncommitted lock state

dmcilvaney:
Add --allow-dirty to build, render, and prepare-sources commands to
control how uncommitted lock file changes are handled during synthetic
history generation.

When --allow-dirty is set, uncommitted lock data is used as a fallback
when no committed lock exists at HEAD, and fingerprint mismatches
produce dirty synthetic entries. When unset, uncommitted state produces
errors with hints suggesting --allow-dirty.

- Add allowDirty + hints plumbing through buildSyntheticCommits
- Extract handleMissingHeadLock and handleDirtyState helpers
- Add Hints() to SourcePreparer interface for caller-facing suggestions
- Always compute fingerprints (local I/O only, no network)
- Add WithPreScript option to scenario ProjectTest
- Scenario build test commits locks before building
- Regenerate CLI docs

### DIFF
--- a/docs/user/reference/cli/azldev_component_build.md
+++ b/docs/user/reference/cli/azldev_component_build.md
@@ -48,6 +48,7 @@ azldev component build [flags]
 
 ```
   -a, --all-components                   Include all components
+  -d, --allow-dirty                      include uncommitted lock file changes in synthetic history
   -p, --component stringArray            Component name pattern
   -g, --component-group stringArray      Component group name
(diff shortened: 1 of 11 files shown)


Tonisal-byte (approved):

dmcilvaney:
Once we get the new changelog feature this is no longer as useful (since majority of packages will show 'dirty' entry correctly. Switching to draft.

PR #190, opened 2026-05-14
fix: balance conditional nesting when removing spec sections

### LINKED ISSUE #144: spec-remove-section / spec-remove-subpackage produce invalid specs when sections are inside %if/%endif
> _Drafted by GitHub Copilot CLI while reviewing PR #132._

`Spec.RemoveSection` and `Spec.RemoveSubpackage` (`internal/rpm/spec/edit.go`) operate on whole-section line ranges and have no awareness of `%if`/`%endif` conditionals. When a section is wrapped in a conditional block, the `%endif` is typically consumed as part of the trailing removed section, while the matching `%if` remains in place. The result is a syntactically invalid spec written to disk; the build then fails far away with a confusing error.

This limitation is honestly documented for both overlays in `docs/user/reference/config/overlays.md` and in the Go doc-comment on `RemoveSubpackage`, but the API still returns `nil` and silently writes a broken spec.

**Reproduction**

```spec
Name: test
Version: 1.0

%description
Main.

%if 0%{?with_devel}
%package devel
Summary: Devel files

%description devel
Devel description.

%files devel
/usr/include/test.h
%endif

%files
/usr/bin/test
```

Applying:

```toml
[[components.test.overlays]]
type = "spec-remove-subpackage"
package = "devel"
```

…produces:

```spec
Name: test
Version: 1.0

%description
Main.

%if 0%{?with_devel}
%files
/usr/bin/test
```

The `%endif` was consumed as part of the trailing `%files devel` section (it sits between that section header and the next section header), and the `%if` is now unbalanced.

**Proposed fix**

`Spec.collectSectionRanges` already runs inside `Visit`, whose context tracks `conditionalDepthChange`. Detect ranges that:

1. Start at non-zero conditional depth, or
2. Have a non-zero net conditional balance change across the range.

In either case, return a clear, actionable error (e.g. *"section %q (package=%q) is wrapped in %if/%endif; remove the conditional first or use spec-search-replace"*) instead of writing invalid output.

Apply uniformly to `RemoveSection` and `RemoveSubpackage`. After the fix, update the limitation prose in `docs/user/reference/config/overlays.md` to describe the new failure mode (clear error at overlay-application time) rather than the silent-corruption mode.

**Workaround until fixed**

Use a `spec-search-replace` overlay to remove the conditional block, or remove the `%if`/`%endif` first via additional overlays before applying the section/sub-package removal.

trungams:
`collectSectionRanges` now adjusts each section range to maintain balanced %if/%endif nesting before removal. This prevents producing invalid specs when sections are adjacent to or wrapped in conditional blocks.

Three cases are handled automatically:
- Balanced conditionals inside a section (no change needed)
- Trailing %if openers belonging to the next section (trimmed)
- Trailing %endif from a wrapping conditional block (trimmed, leaving an empty but valid conditional wrapper)

Conditionals interleaved with section content (spanning case) are detected and reported with an [ErrConditionalSpansSections] error.

Fixes #144

### DIFF
--- a/internal/rpm/spec/edit.go
+++ b/internal/rpm/spec/edit.go
@@ -21,6 +21,11 @@ var ErrNoSuchTag = errors.New("no such tag")
 // ErrSectionNotFound is returned when a requested section does not exist in the spec.
 var ErrSectionNotFound = errors.New("section not found")
 
+// ErrConditionalSpansSections is returned when a conditional block (%if/%endif) spans
+// across section boundaries, making it impossible to safely remove the section without
+// breaking the conditional nesting structure.
+var ErrConditionalSpansSections = errors.New("conditional block spans across section boundaries")
+
 // ErrPatternNotFound is returned when a search pattern does not match any content in the spec.
 var ErrPatternNotFound = errors.New("pattern not found")
 
@@ -195,8 +200,10 @@ func tagFamily(tag string) string {
 	return lower[:end]
 }
 
-// conditionalDepthChange returns +1 for lines that open a conditional block (%if, %ifarch, etc.),
+// conditionalDepthChange returns +1 for lines that open a conditional block,
 // -1 for %endif, and 0 for everything else. Comments are ignored.
+//
+// The recognized conditional openers are: %if, %ifarch, %ifnarch, %ifos, %ifnos.
 func conditionalDepthChange(rawLine string) int {
 	trimmed := strings.TrimSpace(rawLine)
 	if strings.HasPrefix(trimmed, "#") {
@@ -209,15 +216,41 @@ func conditionalDepthChange(rawLine string) int {
 	}
 
 	lower := strings.ToLower(token[0])
-	if lower == "%endif" {
+
+	switch lower {
+	case "%endif":
 		return -1
-	}
-
-	if strings.HasPrefix(lower, "%if") {
+	case "%if", "%ifarch", "%ifnarch", "%ifos", "%ifnos":
 		return 1
+	default:
+		return 0
+	}
+}
+
+// isConditionalBranchDirective returns true for lines that are branch directives
+// within a conditional block. These do not change nesting depth but mark branch
+// boundaries within an enclosing %if/%endif pair. Comments are ignored.
+//
+// The recognized branch directives are: %else, %elif, %elifarch, %elifnarch, %elifos, %elifnos.
+func isConditionalBranchDirective(rawLine string) bool {
+	trimmed := strings.TrimSpace(rawLine)
+	if strings.HasPrefix(trimmed, "#") {
+		return false
 	}
 
-	return 0
+	tokens := strings.Fields(trimmed)
+	if len(tokens) == 0 {
+		return false
+	}
+
+	lower := strings.ToLower(tokens[0])
+
+	switch lower {
+	case "%else", "%elif", "%elifarch", "%elifnarch", "%elifos", "%elifnos":
+		return true
+	default:
+		return false
+	}
 }
 
 // InsertTag inserts a tag into the spec, placing it after the last existing tag from the
@@ -792,11 +825,12 @@ func (s *Spec) RemoveSection(sectionName, packageName string) error {
 // whichever form the spec uses. Specs that mix both forms for the same sub-package
 // (uncommon but legal) require a call per form.
 //
-// Limitation: this method operates on whole-section line ranges and does not understand
-// `%if`/`%endif` conditionals. Sub-packages wrapped in a conditional block will leave
-// an unbalanced `%if` (the `%endif` is typically consumed as part of the trailing
-// sub-package section). Callers that need to handle conditionalized sub-packages must
-// adjust the conditionals themselves.
+// Conditional handling: section ranges are automatically trimmed to maintain balanced
+// `%if`/`%endif` nesting. Sections wrapped in a conditional block will have trailing
+// `%endif` lines excluded from the removal, leaving an empty (but valid) conditional
+// wrapper. Trailing `%if` lines that belong to the next section are similarly excluded.
+// If a conditional block is interleaved with section content in a way that cannot be
+// resolved by trimming, an [ErrConditionalSpansSections] error is returned.
 func (s *Spec) RemoveSubpackage(packageName string) error {
 	slog.Debug("Removing sub-package from spec", "package", packageName)
 
@@ -831,6 +865,13 @@ type sectionLineRange struct {
 // collectSectionRanges walks the spec and returns one [sectionLineRange] for every
 // section whose `(sectName, packageName)` pair satisfies the predicate, in the order
 // they appear in the spec.
+//
+// Each returned range is adjusted to maintain conditional balance: if a range would
+// include trailing `%if` or `%endif` lines that create a nesting imbalance, those
+// lines are trimmed from the range so that removing the range does not break the
+// spec's conditional structure. If a conditional block is interleaved with section
+// content in a way that cannot be resolved by trimming, an [ErrConditionalSpansSections]
+// error is returned.
 func (s *Spec) collectSectionRanges(
 	matches func(sectName, packageName string) bool,
 ) ([]sectionLineRange, error) {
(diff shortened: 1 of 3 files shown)


internal/rpm/spec/edit.go
```diff
@@ -220,6 +225,26 @@ func conditionalDepthChange(rawLine string) int {
...
+	}
+
+	tokens := strings.Fields(trimmed)
+	if len(tokens) == 0 {
+		return false
+	}
+
+	lower := strings.ToLower(tokens[0])
+
+	return lower == "%else" || strings.HasPrefix(lower, "%elif")
```
reubeno: Isn't this too broad? Will this match on `%elifsomething`?

internal/rpm/spec/edit.go
```diff
@@ -866,9 +899,210 @@ func (s *Spec) collectSectionRanges(
...
+// pairs using a stack. Nested pairs are properly matched. Returns an error if there
+// are unmatched `%if` or `%endif` directives.
+func collectConditionalPairs(rawLines []string) ([]conditionalPair, error) {
+	var (
+		pairs []conditionalPair
+		stack []int
+	)
+
+	for lineNum, line := range rawLines {
+		switch conditionalDepthChange(line) {
```
reubeno, line 951: I know the code was preexisting, but can you check if `conditionalDepthChange()` suffers from the same risk of being too broad (e.g., matching `%ifsomething`)?

internal/rpm/spec/edit.go
```diff
@@ -831,6 +857,13 @@ type sectionLineRange struct {
...
 // section whose `(sectName, packageName)` pair satisfies the predicate, in the order
 // they appear in the spec.
+//
+// Each returned range is adjusted to maintain conditional balance: if a range would
+// include trailing `%if` or `%endif` lines that create a nesting imbalance, those
+// lines are trimmed from the range so that removing the range does not break the
+// spec's conditional structure. If a conditional block is interleaved with section
+// content in a way that cannot be resolved by trimming, an [ErrConditionalSpansSections]
+// error is returned.
 func (s *Spec) collectSectionRanges(
```
reubeno, line 875: Can you tell if `spec-append-lines` et al. suffer from the same problem? I'm assuming we may need to update them to use this code path too. Otherwise, the lines might get appended after an %ifdef that wraps a succeeding section.

IIRC, that was one of the reasons that Tobias had added `spec-insert-tag` as a workaround.

I'm okay if you want to split that off as a separate PR, but since you're already in this code path / logic would be great if you could sort that out too.

internal/rpm/spec/edit.go
```diff
@@ -866,9 +899,210 @@ func (s *Spec) collectSectionRanges(
...
+		}
+
+		if insideLine < trimmed {
+			trimmed = insideLine
+		}
+
+		// If the straddling line is an %if (opener inside, closer outside),
+		// check for real content between the %if and the range end. Such content
+		// would belong to this section but span into the next via the conditional.
+		if ifInside {
```
reubeno, line 1019: Maybe I'm missing something but don't we want to use the same logic here when `%endif` was inside but its starting `%if` wasn't? I'd only expect that to be okay if there's no content after.

internal/rpm/spec/edit.go
```diff
@@ -866,9 +899,210 @@ func (s *Spec) collectSectionRanges(
...
+	// we'd be leaving behind part of the section the caller asked to remove.
+	if err := validateNoContentAfter(trimmed-1, sectionRange.end, rawLines); err != nil {
+		return sectionRange, fmt.Errorf(
+			"section at lines %d-%d has a conditional block that spans into the next section; "+
+				"use a spec-search-replace overlay to adjust conditionals before removing:\n%w",
+			sectionRange.start+1, sectionRange.end, ErrConditionalSpansSections,
+		)
+	}
+
+	// Trim trailing blank/comment lines for clean output.
```
reubeno: I actually do like the trimming, but for consistency's sake can we keep this trimming consistent across both the case where we need to rebalance and we don't? (My preference would be to do it in both places.)

internal/rpm/spec/edit.go
trungams: This is already handled later in the function, on line 1035

internal/rpm/spec/edit.go
trungams: This should cover `%elif/%elifarch/%elifnarch`. I think other keywords with this same `elif` prefix should carry the same semantics, at least for the current version of spec files

internal/rpm/spec/edit.go
trungams: Similar answer as above. I think prefix check for this is fine

internal/rpm/spec/edit.go
trungams: I think it does, since the limitation initially comes from `Spec.Visit` not accounting for conditionals in section ranges. I need a bit more time to think of a direct improvement to `Spec.Visit`, so I'd prefer to do it in a follow-up PR. The solution I'm adding here is meant to go on top of `collectSectionRanges` to have as little impact on existing spec files as possible.

internal/rpm/spec/edit.go
reubeno: > This is already handled later in the function, on line 1035

Understood, but isn't the conditional unneeded complexity? Or is there some other case I'm not thinking of?

internal/rpm/spec/edit.go
reubeno: Fair point. I'd like to see you continue working in this area after this PR -- to clean it up a bit. If we get more tests in place, we can have the confidence to better unify/factor the logic so it doesn't continue to accrete complexity.

internal/rpm/spec/edit.go
reubeno: But there's nothing stopping a spec from defining a macro called %elifsomething and us incorrectly considering it a conditional -- and from a grammar perspective, there's no way to tell if it's supposed to be a special conditional or a plain macro expansion.

It's an edge case, but it seems like it should be simple to address by making the check stricter.

internal/rpm/spec/edit.go
reubeno: Is that because you think it's sufficiently unlikely that you don't think the complexity is worth it, or because there's something that would prevent it from being an issue?

internal/rpm/spec/edit.go
trungams: That's a good point, and yes it's simple enough. I'll go update it

internal/rpm/spec/edit.go
trungams: I'm trying to cover the general cases so yes, it's very unlikely this happens and implies a spec is badly written, so we need to throw an error.

internal/rpm/spec/edit.go
trungams: I'm going to add a post-processing step that cleans up the specs after all overlays are applied

internal/rpm/spec/edit.go
trungams: updated

internal/rpm/spec/edit.go
trungams: I didn't think of the macro scenario. I updated the code to check more strictly

internal/app/azldev/core/sources/overlays.go
```diff
@@ -35,6 +35,39 @@ func isSpecFile(filePath string) bool {
...
+		return fmt.Errorf("failed to open spec %#q for writing:\n%w", specPath, err)
+	}
+
+	defer defers.HandleDeferError(specFile.Close, &err)
+
+	err = openedSpec.Serialize(specFile)
+	if err != nil {
+		return fmt.Errorf("failed to serialize spec %#q:\n%w", specPath, err)
+	}
+
```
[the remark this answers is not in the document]
trungams: It's unlikely we'd need more than 3 consecutive blank lines in a spec for anything meaningful

internal/rpm/spec/edit.go
reubeno: What kind of post processing?

internal/app/azldev/core/sources/overlays.go
```diff
@@ -35,6 +35,39 @@ func isSpecFile(filePath string) bool {
 	return strings.HasSuffix(filePath, ".spec")
 }
 
+// collapseSpecBlankLines opens a spec file, collapses consecutive blank lines, and
+// writes it back. This is a post-processing step intended to clean up gaps left by
+// section or tag removal overlays.
+func collapseSpecBlankLines(fs opctx.FS, specPath string) error {
```
reubeno: This will introduce more deviations from input upstream specs than strictly needed; is this necessary? I don't think we should be in the business of rewriting the specs for readability, unless it's scoped to the specific overlay changes we're making.

 I think the "trim" you had been doing for section collection was fine because it didn't actually remove any content, it just considered the section to end a bit earlier so that appended lines / removed sections preserve the space between the sections.

(Minor note: it's also quite inconsistent that this logic gets applied unilaterally to all parts of the spec, including sections not overlaid -- but is only triggered if a certain subset of overlays are requested.)

internal/rpm/spec/edit.go
trungams: ~~I re-read the code and realized "for clean output" was a bit misleading and that trimming logic didn't actually do anything useful, so I removed it. Output should look consistent now~~

I think it's better to remove the trimming at least for this PR, it's sending me down a rabbit hole trying to properly define section boundaries and accounting for comments/whitespaces. I'll try to keep this PR simple for now

internal/app/azldev/core/sources/overlays.go
trungams: I was overthinking it - 2am thoughts :(

reubeno: No worries. I think there's a compromise somewhere in there that's reasonable -- and fine for a subsequent PR.

reubeno:
Changes are looking good; primarily waiting on https://github.com/microsoft/azurelinux/pull/17259 to go in first so this won't automatically be pulled in. (Looks like that would result in 2 specs rendering differently and resulting errors. We'll want to ingest this change and update those 2 rendered specs in a single PR to `microsoft/azurelinux`.)

PR #197, opened 2026-05-18
test: Added end-to-end upstream render test

Tonisal-byte:
Add scenario tests for the synthetic dist-git render pipeline. Existing render tests never committed lock files, so buildSyntheticCommits always returned early (no lock at HEAD) and the synthetic history codepath was never exercised end-to-end.

New tests:

`TestRenderUpstreamComponentWithSyntheticHistory`: Renders an upstream Fedora component (lolcat) with a pre-baked lock file pinned to a known dist-git commit. Validates the full pipeline: upstream fetch, overlay application, synthetic commit generation, and rpmautospec/changelog expansion.

`TestRenderLocalSpecWithSyntheticHistory`: Renders a local %autorelease spec with a pre-baked lock file. Verifies that synthetic commits are created from lock file fingerprint changes and rpmautospec processes the resulting git history.

`TestRenderLocalSpecWithDirtyDetection`: Renders a local %autorelease spec with a stale lock file fingerprint and an overlay that changes the runtime fingerprint. Validates that dirty detection fires (uncommitted fingerprint != HEAD lock fingerprint), the overlay is applied, and rpmautospec processes the spec with the dirty synthetic commit.

### DIFF
--- a/internal/providers/sourceproviders/fedorasource/fedorasource.go
+++ b/internal/providers/sourceproviders/fedorasource/fedorasource.go
@@ -470,16 +470,23 @@ const (
 )
 
 // validateAbsoluteURL parses uri and verifies it is an absolute URL with a
-// non-empty scheme and host. The label parameter is used in error messages to
-// identify the URL's purpose (e.g. "lookaside", "dist-git").
+// non-empty scheme. For schemes that require a host (everything except
+// "file"), a non-empty host is also required. The label parameter is used
+// in error messages to identify the URL's purpose (e.g. "lookaside",
+// "dist-git").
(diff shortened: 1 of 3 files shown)


Tonisal-byte:
Please take a look when you have the time @dmcilvaney

reubeno:
@Tonisal-byte Is there a way to validate the actual final `Release` and changelog values by controlling the "input dist-git" repo being cloned from? (Perhaps in a separate test that redirect dist-git to a local test git repo we create just for the test?) Or would you see that as an additive test beyond this?

reubeno (approved):

PR #201, opened 2026-05-20
test(e2e): add end-to-end scenario tests against real upstream repos

reubeno:
Introduces a new `mage e2e` target and `//go:build e2e` suite that exercises `azldev component` commands against actual Azure Linux source repositories.

These tests are excluded from `mage scenario`/`mage all` and intended for CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,3 +62,38 @@ jobs:
       - name: Run tests
         run: |
           mage -v scenario
+
+  e2e-tests:
+    name: "End-to-end tests"
+    # E2E tests are tagged separately from scenario tests (build tag 'e2e' vs.
+    # 'scenario') so they do not run as part of 'mage scenario' / 'mage all'.
+    # They clone large upstream repositories and exercise full mock pipelines,
+    # so they live in their own job and are gated on network access.
+    strategy:
+      matrix:
+        host: [ubuntu-latest]
+    runs-on: ${{ matrix.host }}
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Get go tools
+        uses: ./.github/getgotools
+
+      - name: Disable unix_chkpwd policy
+        run: |
+          # We disable the unix-chkpwd AppArmor profile to avoid errors running
+          # containerized tests. We also need to purge a few packages to avoid
+          # profile conflicts.
+          set -euxo pipefail
+          sudo apt-get update
+          sudo apt-get purge firefox passt
+          sudo systemctl reload apparmor.service
+          sudo apt-get install apparmor-utils
+          sudo aa-disable /usr/sbin/unix_chkpwd
+
+      - name: Run tests
+        run: |
+          mage -v e2e
--- a/magefiles/magescenario/magescenario.go
+++ b/magefiles/magescenario/magescenario.go
@@ -43,10 +43,82 @@ func ScenarioUpdate() error {
 	return nil
 }
 
+// E2E runs the end-to-end tests. These are tagged with the `e2e` build tag
+// (separate from `scenario`) so that they do not run as part of `mage scenario`,
+// `mage scenarioUpdate`, or `mage all` — they are intentionally heavier
+// (network access, large clones, full mock pipelines) and meant to be run
+// from CI on a dedicated job.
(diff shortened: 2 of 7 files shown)


scenario/component_e2e_test.go
```diff
@@ -0,0 +1,224 @@
...
+// repository at [azureLinuxBranch] inside a privileged, networked scenario
+// container, runs `azldev <azldevArgs...>` from the clone root, and asserts
+// that the working tree is clean afterwards (no modified, deleted, or
+// untracked files).
+//
+// The script also performs lightweight pre-flight sanity checks against the
+// upstream layout, prints the resolved upstream commit SHA for traceability,
+// and emits a verbose diagnostic block (full status, diff, untracked listing)
+// before exiting non-zero on failure. This makes the difference between an
+// azldev regression and an upstream-branch drift easy to spot in CI logs.
```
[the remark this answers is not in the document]
Tonisal-byte: I think this could be a good addition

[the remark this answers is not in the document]
reubeno: The main point of this test is an early warning sign against the *latest* commit in microsoft/azurelinux, as it's the best indication of whether azldev changes will break it when they're ingested.

I understand the concerns around a floating dependency, but this is by design from my perspective.

.github/workflows/test.yml
```diff
@@ -62,3 +62,38 @@ jobs:
...
+          set -euxo pipefail
+          sudo apt-get update
+          sudo apt-get purge firefox passt
+          sudo systemctl reload apparmor.service
+          sudo apt-get install apparmor-utils
+          sudo aa-disable /usr/sbin/unix_chkpwd
+
+      - name: Run tests
+        run: |
+          mage -v e2e
```
Tonisal-byte, line 99: Kind of related to the other comment, I looked to see why the tests currently fail. I believe at the time this scenario test was triggered we had a broken `qemu` in our 4.0 branch. This render failure caused the test to fail. I don't know how ideal it would be having this test block merging when it takes a while to complete and could fail from upstream hiccups. I see the value of having this test but could we leave it as non-blocking (possibly a bad idea)

magefiles/magescenario/magescenario.go
```diff
@@ -36,14 +38,82 @@ func Scenario() error {
...
+
+	mageutil.MagePrintf(mageutil.MsgSuccess, "End-to-end tests complete: %v.\n", time.Since(startTime))
+
+	return nil
+}
+
 func mageScenarioCommon(doSnapshotUpdate bool) error {
+	options := scenarioGoTestOptions{
+		buildTag:         "scenario",
+		timeout:          "60m",
```
Tonisal-byte, line 83: Is it possible this 60m limit gets exhausted in the medium-term future? (Our current e2e -a executes in ~40m)

magefiles/magescenario/magescenario.go
```diff
@@ -25,7 +25,9 @@ func Scenario() error {
 
 	mageutil.MagePrintln(mageutil.MsgStart, "Starting scenario tests...")
 
-	mg.SerialDeps(mg.F(mageScenarioCommon, false))
+	if err := mageScenarioCommon(false); err != nil {
```
dmcilvaney: Out of curiosity, was this causing issues?

dmcilvaney (approved):

magefiles/magescenario/magescenario.go
reubeno: That's a good call. I'll look at maybe increasing?

.github/workflows/test.yml
reubeno: This is a test that I actually would consider making non-blocking. The way I see it, it's an early warning sign that azldev changes could cause breaks or required re-renders in azurelinux.  I'd rather know that upfront than when we get around to updating azldev on the azurelinux side.

In the case of qemu, it was by design. We merged a change to azldev that changed the behavior of section removal, but in a way that we believed would be safe and better for ifdef cases. But it still brought re-render requirements nonetheless.

magefiles/magescenario/magescenario.go
dmcilvaney: Render is around 45 minutes on the default github runners, maybe double to be safe?

magefiles/magescenario/magescenario.go
reubeno: Good question; copilot thought it was, but I looked at it again and didn't see a need to make the change. I've reverted it to what it was before.

PR #204, opened 2026-05-21
feat(component query): bulk spec querying with --arch enforcement

liunan-ms:
Introduces `azldev component query` for bulk-querying rendered RPM specs inside a single shared mock chroot, with arch-aware policy enforcement.
### CLI
- New `azldev component query` command — selects components via the standard component filter, renders results as a table / CSV / Markdown / JSON.
- New `--arch` flag (default `x86_64`, also accepts `aarch64`) driving `rpmspec --target=<arch>` and ExclusiveArch/ExcludeArch policy.

### Docs
CLI reference under `docs/user/reference/cli/` regenerated.


internal/app/azldev/core/sources/specquery.go
```diff
@@ -0,0 +1,258 @@
...
+		return fmt.Errorf("spec path %#q must be relative", relPath)
+	}
+
+	cleaned := filepath.Clean(relPath)
+	if cleaned != relPath {
+		return fmt.Errorf("spec path %#q must be in canonical form", relPath)
+	}
+
+	if strings.Contains(cleaned, "..") {
+		return fmt.Errorf("spec path %#q must not contain path traversal", relPath)
```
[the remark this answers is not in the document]
liunan-ms: Such case doesn't appear in practice. validateSpecRelPath only receives spec paths, shaped like <bucket>/<name>/<name>.spec. The bucket is a single lowercase letter; the name has already been validated by fileutils.ValidateFilename, which rejects `..` and other path-tricks. So a foo..bar/spec.spec input can never reach this function via legitimate flows.

scenario/testdata/defaultconfigs/distros/azl/azl.distro.toml
```diff
@@ -19,7 +19,7 @@ mock-config-x86_64 = "mock/azurelinux-4-x86_64.cfg"
 mock-config-aarch64 = "mock/azurelinux-4-aarch64.cfg"
 
 [distros.azurelinux.versions.'4.0'.default-component-config]
-spec = { type = "upstream", upstream-distro = { name = "fedora", version = "42" } }
+spec = { type = "upstream", upstream-distro = { name = "fedora", version = "43" } }
```
liunan-ms, line 22: `java-srpm-macros` is not available in f42 as a standalone binary RPM. Bumping the test mock to releasever=43 (and the matching f43-build URL) to unblock the scenario tests.

reubeno:
@dmcilvaney would you mind reviewing this? You'll be most familiar with the parts that touch the parallel mock processor.

internal/app/azldev/core/sources/specquery.go
```diff
@@ -0,0 +1,258 @@
...
+
+		seen[input.Name] = true
+	}
+
+	return nil
+}
+
+// validateSpecRelPath rejects spec relative paths that could escape the
+// specs-dir bind mount or contain control characters.
+func validateSpecRelPath(relPath string) error {
```
dmcilvaney: We should consolidate path validation to one spot, we have some semi common functions in `fileutils.ValidateFilename`, should augment that if it has holes.

internal/app/azldev/command.go
```diff
@@ -124,13 +124,40 @@ func runFuncInternal(innerFunc CmdWithExtraArgsFuncType, requireConfig bool) cob
 				command.SilenceUsage = false
 			}
 
+			// Inner funcs may return partial results alongside an error
+			// (e.g. some items succeeded, some failed). Render what we have
+			// before propagating so callers see the partial output and the
+			// process still exits non-zero.
+			if results != nil && !isNilValue(results) {
```
dmcilvaney, line 164: Can you update render to also support this now? I worked around it by returning no error even when things failed. If we can still output the results table even on failure, I think returning an error code does make sense. Might need to update `checkRenderErrors()` if I recall right.

This might need an associated fix in the main repo's PR checks, but we can do that when we move the alzdev version pin forward.

dmcilvaney, line 164: Same deal with `update`, although there I think we just want to call `filterDisplayResults()` before returning and add `results[idx].Error != ""` along with the other criteria. Otherwise I think it will dump 7k+ values out on an error.

internal/app/azldev/core/sources/query_process.py
```diff
@@ -0,0 +1,444 @@
...
+            "name": name,
+            "srpmOut": "",
+            "binOut": "",
+            "error": f"spec file not found: {comp['specRelPath']}",
+        }
+
+    # Apply per-spec rewrites (e.g. ghc.spec) to a scratch copy if needed.
+    # _sourcedir/_specdir stay pinned to the original spec's directory via
+    # _rpmspec_args, so sidecar files still resolve correctly.
+    effective_spec = _maybe_rewrite_spec(spec_path, scratch_dir, name)
```
dmcilvaney, line 345: Looks like this doesn't take the new `effective_spec` directory to use for spec_dir, is that an issue, or intended? (ie we want a different spec, but to still read everything as-if the spec was in its old spot).

scenario/internal/projecttest/testspec.go
```diff
@@ -15,10 +15,11 @@ const NoArch = "noarch"
...
 // TestSpec represents an RPM spec being composed for testing purposes.
 type TestSpec struct {
-	name      string
-	version   string
-	release   string
-	buildArch string
+	name        string
+	version     string
+	release     string
+	buildArch   string
```
dmcilvaney: Seems like the arch logic is mostly untested end-to-end. Since a lot of it lives in the python side might be easiest to test here via scenario test.

internal/app/azldev/core/sources/query_process.py
```diff
@@ -0,0 +1,444 @@
...
+    """
+    rewrites = _SPEC_REWRITES.get(os.path.basename(spec_path))
+    if not rewrites:
+        return spec_path
+
+    with open(spec_path, encoding="utf-8", errors="replace") as src:
+        content = src.read()
+
+    for find, replace in rewrites:
+        content = content.replace(find, replace)
```
dmcilvaney, line 175: I've had to do the same find/replace pattern when doing similar things, I haven't found a better way either, but we should consider erroring if the string isn't found anymore because the upstream spec changed or got overlays added, might make it easier to fix than a random parse error thats hard to diagnose.

internal/rpm/specquery.go
```diff
@@ -26,6 +26,10 @@ type SpecInfo struct {
 	Name          string
 	Version       Version
 	RequiredFiles []string
+	// Subpackages lists the binary package names the spec produces, in the
+	// order rpmspec reports them. Empty when not queried (e.g., the
+	// per-component buildenv path that only requests --srpm output).
+	Subpackages []string
```
dmcilvaney, line 32: Nit: Are we worried about cases were a package produces no outputs? Maybe make it nil-able to differentiate?

internal/app/azldev/command.go
dmcilvaney, line 164: Nit: Maybe update the comments for the various `CmdFuncType` public functions to indicate this behavior. It may be a bit surprising that a failed command still outputs results.

internal/app/azldev/core/sources/query_process.py
```diff
@@ -0,0 +1,444 @@
...
+        effective_spec,
+        comp["subpackagesQueryFormat"],
+        False,
+        with_,
+        without,
+        defines,
+        arch,
+    )
+    # Insert --builtrpms right after `-q` so it associates with the query.
+    bin_args.insert(2, "--builtrpms")
```
dmcilvaney: Could use something like `bin_args.index("-q") + 1` to find the -q so its stable if args move around.

internal/app/azldev/core/sources/query_process.py
```diff
@@ -0,0 +1,444 @@
...
+    "amd64": "x86_64",
+    "arm64": "aarch64",
+}
+
+
+def _canonicalize_arch_token(token):
+    return _ARCH_ALIASES.get(token, token)
+
+
+def _is_arch_excluded(arch, exclusive_arch, exclude_arch):
```
dmcilvaney: 50/50 if this should live in the golang side somehow, and get passed in as a table. I'm not aware of any other go side features that will care about arch parsing, so I think its fine here, but that might change.

internal/app/azldev/core/sources/specquery.go
liunan-ms: Added a ValidateRelPath to fileutils (built on top of ValidateFilename per segment) and delegate validateSpecRelPath to it.

internal/rpm/specquery.go
liunan-ms: Good question, I don't think we need to. The batched query path always populates Subpackages (even to an empty slice) on a successful query, so len == 0 already means "queried, spec produces no built binary RPMs." The only caller that leaves it nil is the legacy per-component QuerySpec, which doesn't run the subpackage query at all. Happy to revisit if a real caller needs the distinction.

internal/app/azldev/core/sources/query_process.py
liunan-ms: Agreed, keeping it here for now since query_process.py is the only consumer. If a Go-side feature ever needs the same canonicalization, the natural refactor is to define the table once in Go (e.g. under internal/utils/qemu) and ship it down via the inputs JSON alongside arch. I'll leave a TODO if you'd like.

internal/app/azldev/core/sources/query_process.py
liunan-ms: Fixed in [efd4560](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/efd456020cec1b52c4ded85c962e577d977f8124)

internal/app/azldev/command.go
liunan-ms: Fixed in [efd4560](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/efd456020cec1b52c4ded85c962e577d977f8124)

scenario/internal/projecttest/testspec.go
liunan-ms: Fixed in [efd4560](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/efd456020cec1b52c4ded85c962e577d977f8124)

internal/app/azldev/core/sources/query_process.py
liunan-ms: Fixed in [efd4560](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/efd456020cec1b52c4ded85c962e577d977f8124)

internal/app/azldev/command.go
liunan-ms: Fixed in [efd4560](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/efd456020cec1b52c4ded85c962e577d977f8124)

liunan-ms: Fixed in [efd4560](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/efd456020cec1b52c4ded85c962e577d977f8124)

internal/app/azldev/command.go
```diff
@@ -124,13 +148,40 @@ func runFuncInternal(innerFunc CmdWithExtraArgsFuncType, requireConfig bool) cob
...
+				if reportErr := reportResults(env, results); reportErr != nil {
+					return errors.Join(err, reportErr)
+				}
+			}
```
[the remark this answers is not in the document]
liunan-ms: fixed in [5dadbf7](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/5dadbf7ca2e1c23428fb8ca7d9d53c68f55c8f5b)

internal/app/azldev/cmds/component/query.go
```diff
@@ -50,43 +71,247 @@ slower than 'list' but more informative.`,
...
+
+	inputs, skipped, err := buildSpecQueryInputs(env, comps.Components(), renderedSpecsDir)
 	if err != nil {
-		return results, fmt.Errorf("failed to resolve components:\n%w", err)
+		return nil, err
+	}
+
+	if len(inputs) == 0 {
+		return nil, fmt.Errorf("no components have a rendered spec on disk; skipped %d", skipped)
 	}
```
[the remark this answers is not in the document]
liunan-ms: fixed in [5dadbf7](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/5dadbf7ca2e1c23428fb8ca7d9d53c68f55c8f5b)

internal/rpm/specquery.go
```diff
@@ -125,6 +129,59 @@ func (q *SpecQuerier) composeRpmspecCmdline(specPath string) (result []string) {
...
+
+		if strings.HasPrefix(trimmed, "error: ") || strings.HasPrefix(trimmed, "warning: ") {
+			slog.Debug("Ignoring rpmspec error", "line", trimmed)
+
+			continue
+		}
+
+		if after, ok := strings.CutPrefix(trimmed, "subpkg="); ok && after != "" {
+			result = append(result, after)
+		}
```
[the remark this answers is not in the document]
liunan-ms: fixed in [5dadbf7](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/5dadbf7ca2e1c23428fb8ca7d9d53c68f55c8f5b)

internal/app/azldev/cmds/component/query.go
```diff
@@ -50,43 +71,247 @@ slower than 'list' but more informative.`,
...
+			slog.Debug("Failed to clean up scratch directory", "path", scratchDir, "error", removeErr)
+		}
+	}()
+
+	archStr := options.Arch.String()
+
+	queryResults, err := mockProcessor.BatchQuerySpecs(
+		env, env, renderedSpecsDir, scratchDir, archStr,
+		inputs, env.FS(), env.CPUBoundConcurrency(),
+	)
```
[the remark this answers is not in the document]
liunan-ms: fixed in [5dadbf7](https://github.com/microsoft/azure-linux-dev-tools/pull/204/commits/5dadbf7ca2e1c23428fb8ca7d9d53c68f55c8f5b)

internal/app/azldev/core/sources/query_process.py
```diff
@@ -0,0 +1,472 @@
...
+        with_,
+        without,
+        defines,
+        arch,
+        source_dir=source_dir,
+    )
+    # Insert --builtrpms right after `-q` so it associates with the query.
+    # Look up `-q` rather than hard-coding the index so this stays correct if
+    # _rpmspec_args ever reorders its preamble.
+    bin_args.insert(bin_args.index("-q") + 1, "--builtrpms")
```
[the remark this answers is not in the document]
liunan-ms: The lookup-by-value already handles the reorder concern, and there's no realistic user input that can spoof -q in our -D key value shape. I'd rather not widen the signature for a single caller.

dmcilvaney (approved):

scenario/component_query_test.go
```diff
@@ -61,4 +67,139 @@ func TestQueryingAComponent(t *testing.T) {
...
+	for _, tc := range cases {
+		t.Run(tc.arch, func(t *testing.T) {
+			t.Parallel()
+
+			results := projecttest.NewProjectTest(
+				project,
+				[]string{"component", "query", "-a", "--arch", tc.arch},
+				projecttest.WithTestDefaultConfigs(),
+				projecttest.WithPreCommand("component", "render", "-a"),
+			).RunInContainer(t)
```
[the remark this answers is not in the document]
liunan-ms: Not actually shared. Each subtest gets its own `t.TempDir()` and isolated container. In `projecttest.RunInContainer` (projecttest.go:98-105) the project value is re-serialized into a fresh per-subtest staging dir, then runs in its own privileged container with its own mock. The shared project value is read-only metadata (specs + config), not on-disk state, so `component render -a` and the subsequent `component query` always operate on the subtest's private filesystem. No write race possible between the parallel arch cases.

dmcilvaney (approved):

PR #212, opened 2026-05-27
feat(cli): add history command to show changes made to components

dmcilvaney:
If a single component is specified, the history command will show the full details of the changes made to that component, including commit details if '-O json' is used.


reubeno:
@dmcilvaney Is your plan to port this atop #217?

dmcilvaney:
> @dmcilvaney Is your plan to port this atop #217?

We could do that, but it wasn't needed for my use-case.

internal/app/azldev/cmds/component/history.go
```diff
@@ -0,0 +1,465 @@
...
+	for _, comp := range comps {
+		name := comp.GetName()
+
+		// Read the raw per-component config (as authored in TOML), not the
+		// resolved one returned by comp.GetConfig() -- the resolver
+		// pre-merges project- and group-level defaults, which would
+		// otherwise look like per-component customizations.
+		var items []CustomizationItem
+		if raw, ok := env.Config().Components[name]; ok {
+			items = collectCustomizations(name, &raw)
```
liunan-ms, line 417: The history command deliberately reads the raw per-component TOML config, not the resolved one. After migrating overlays from comp.toml files to  overlay.toml files, the customizations will not be included in the `azldev comp hist` output as the overlays are discovered via a project-wide default added to comp.toml. Is the history command meant to keep this isolation (missing customizations) or should we merge the resolved overlays into the raw config?

liunan-ms (commented):
The history command would be very helpful! I just have a question about reading raw config or resolved config for customizations.

internal/app/azldev/cmds/component/history.go
dmcilvaney: Hmm... good point, this predates that. I'm not actually sure here, maybe the easy answer is just accept the global changes on all results.

liunan-ms: I'm fine with the current implementation. If we need to add the customizations in the history output, it'd be helpful to add the commit message on each overlays as well.

liunan-ms (approved):
LGTM

PR #213, opened 2026-05-28
feat(repo): add 'azldev repo query' for inspecting and managing RPM repositories

liunan-ms:
This PR adds a new repo command group with a query subcommand that wraps dnf, auto-discovers the RPM sub-repos it should enable, probes them for reachability, then exec's into dnf with the surviving slots wired up via --repofrompath / --enablerepo. `--no-debuginfo` / `--no-srpms` are available for dropping sub-repos by their declared kind.

Two selection modes:

1. URL mode (azldev repo query --repo-prefix URL [--template] …)
Each URL is expanded against an rpm-repo-set-template (default `--template`) into one sub-repo per template row, fanned out per `--arch` where the row's subpath contains `$basearch` `http://`, `https://`, and `file://` prefixes are all accepted, so a locally synthesized repo tree (e.g. one produced by a build) can be queried directly without serving it over HTTP

2. Project-config mode (azldev repo query --version VER [--use-case rpm-build|image-build] ...)
Resolves the inputs list of the default distro's VER version from [distros.<d>.versions.<VER>.inputs]. GPG keys, per-repo arch allowlists, and metalink-only rejection come from [resources.rpm-repo-sets.*] / [resources.rpm-repos.*]. `--use-case` defaults to rpm-build. Mutually exclusive with `--repo-prefix` / `--template`.

Example output:
1. URL mode
Screenshot 2026-06-01 140551
2. Project-config mode using the distro version inputs in [azurelinux.repos.distro.toml](https://github.com/microsoft/azurelinux/pull/17597)
image


defaultconfigs/defaultconfigs.go
```diff
@@ -42,3 +42,16 @@ func CopyTo(dryRunnable opctx.DryRunnable, fs opctx.FS, destPath string) (rootCo
 
 	return filepath.Join(destPath, rootDefaultConfigFilename), nil
 }
+
+// ReadRootConfig returns the bytes of the embedded root default configuration
+// file (defaults.toml). Callers that need to inspect a single value from the
+// defaults without paying for the full project-config loader can decode the
```
reubeno: Is this just a performance optimization? I don't know that we want to set the precedent for directly inspecting the configs without going through validation, include loading, group inheritance, etc.

internal/app/azldev/cmds/repo/repo.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev"
+	"github.com/spf13/cobra"
+)
+
+// OnAppInit registers the `repo` command tree with app.
+func OnAppInit(app *azldev.App) {
+	cmd := &cobra.Command{
+		Use:   "repo",
+		Short: "Inspect and synthesize Azure Linux RPM repositories",
```
reubeno: This is relatively specifc (i.e., "inspect and synthesize"). I think the command tree could be more broadly described as inspecting and managing RPM repositories. They need not even be Azure Linux, as we could use it to query, say, Fedora too.

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,329 @@
+// Copyright (c) Microsoft Corporation.
```
reubeno, line 1: Meta-question: for querying, is there a reason we can't just wrap `dnf`? The value is in being able to wire up all the required repo flags, gpg options, etc. -- but there doesn't seem like there's much value in doing the heavy lifting of parsing and inspecting rpm repo metadata.

The current dnf wrapper script is something I use often -- and it lets me directly use all dnf commands  (e.g., `repoclosure`, `repoquery`, `list`, etc.) as well as all their standard options.

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,329 @@
...
+	cmd.Flags().StringVar(&options.Template, "template", repolayout.DefaultTemplateName,
+		"name of the rpm-repo-set-template to expand each --repo-prefix against")
+	cmd.Flags().StringSliceVar(&options.Arches, "arch", repolayout.DefaultArches,
+		"comma-separated arches to expand $basearch over")
+	cmd.Flags().StringVar(&options.OutputDir, "output-dir", "",
+		"with --output json, write per-slot packages.json files under <dir>/<template>/<subrepo>/[<arch>/]")
+	cmd.Flags().BoolVar(&options.Synthesize, "synthesize", false,
+		"reserved for a future release")
+
+	if err := cmd.MarkFlagRequired("repo-prefix"); err != nil {
```
reubeno: Why would this be required if the user wants to point to an RPM set?

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,329 @@
...
+
+	cmd.Flags().StringArrayVar(&options.RepoPrefixes, "repo-prefix", nil,
+		"layout prefix; expanded under the chosen --template; may be repeated")
+	cmd.Flags().StringVar(&options.Template, "template", repolayout.DefaultTemplateName,
+		"name of the rpm-repo-set-template to expand each --repo-prefix against")
+	cmd.Flags().StringSliceVar(&options.Arches, "arch", repolayout.DefaultArches,
+		"comma-separated arches to expand $basearch over")
+	cmd.Flags().StringVar(&options.OutputDir, "output-dir", "",
+		"with --output json, write per-slot packages.json files under <dir>/<template>/<subrepo>/[<arch>/]")
+	cmd.Flags().BoolVar(&options.Synthesize, "synthesize", false,
```
reubeno: Let's leave out future options for latest PRs?

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,329 @@
...
+<dir>/<template>/<subrepo-name>/[<arch>/]packages.json and the command itself
+emits a summary table.`,
+		RunE: azldev.RunFuncWithoutRequiredConfigWithExtraArgs(func(env *azldev.Env, _ []string) (interface{}, error) {
+			return RunQuery(env, &options)
+		}),
+	}
+
+	cmd.Flags().StringArrayVar(&options.RepoPrefixes, "repo-prefix", nil,
+		"layout prefix; expanded under the chosen --template; may be repeated")
+	cmd.Flags().StringVar(&options.Template, "template", repolayout.DefaultTemplateName,
```
reubeno: Is it possible to have reasonable defaults that leverage the default RPM inputs in the currently selected distro? My thought is that we'll have distro "versions" for "koji" vs. "dev" vs. "prod" -- and distro version selector may allow simplifying the querying?

reubeno (commented):
Thanks for putting this together! A few questions -- happy to chat synchronously to go over some of it.

internal/app/azldev/cmds/repo/query.go
liunan-ms: Yes, this's added into the new `--version` mode in this PR. It pulls the repo list straight from [distros.<default-distro>.versions.<VER>.inputs] of the project's selected distro, only version is needed.

internal/app/azldev/cmds/repo/query.go
liunan-ms: synthesize support is removed from the current PR and will be added in a separate PR.

internal/app/azldev/cmds/repo/query.go
liunan-ms: "--repo-prefix" now is not a required flag, repeatable repo prefixes are allowed for URL mode.

internal/app/azldev/cmds/repo/query.go
liunan-ms: Refactored the code to make repo query command a dnf wrapper.

defaultconfigs/defaultconfigs.go
liunan-ms: Removed this helper. Refactored the code to use project-config with full validation.

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+)
+
+// DefaultTemplateName is the name of the built-in standard Azure Linux layout
+// template defined in `defaultconfigs/content/defaults.toml`.
+const DefaultTemplateName = "azl-standard"
```
reubeno: Since this is defined by defaultconfigs, could we define this const in the defaultconfigs package too? That would leave them close together.

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+	// SubrepoName is the [projectconfig.SubrepoSpec.Name] this repo was expanded from.
+	SubrepoName string
+	// Kind mirrors [projectconfig.SubrepoSpec.Kind], defaulted.
+	Kind projectconfig.SubrepoKind
+	// Arch is the substituted `$basearch` value, or "" when the subpath has no
+	// `$basearch` (e.g., a single source/SRPM repo).
+	Arch string
+	// URL is the fully-resolved repo base URL.
+	URL string
+	// PrefixIndex is the 1-based position of the originating --repo-prefix among
```
reubeno: Why are these prefix indices/counts needed? It seems like a bit of knowledge from above leaking down into this layer.

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+		kind := sub.Kind.Default()
+
+		if strings.Contains(sub.Subpath, basearchPlaceholder) {
+			for _, arch := range arches {
+				out = append(out, InputRepo{
+					TemplateName: templateName,
+					SubrepoName:  sub.Name,
+					Kind:         kind,
+					Arch:         arch,
+					URL:          base + "/" + strings.ReplaceAll(sub.Subpath, basearchPlaceholder, arch),
```
reubeno: Can we use `URL.JoinPath` instead here?

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+
+			continue
+		}
+
+		out = append(out, InputRepo{
+			TemplateName: templateName,
+			SubrepoName:  sub.Name,
+			Kind:         kind,
+			URL:          base + "/" + sub.Subpath,
```
reubeno: Same as above.

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+		})
+	}
+
+	return out
+}
+
+// DedupInputRepos drops duplicate entries by URL while preserving order.
+func DedupInputRepos(repos []InputRepo) []InputRepo {
```
reubeno, line 113: Does the `lo` package offer any helpers that can do this for us? Perhaps `lo.Uniq` or perhaps they have a disjoint set data structure we could use?

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+
+		out = append(out, repo)
+	}
+
+	return out
+}
+
+// NormalizePrefix validates p as an http://, https://, or file:// URL and
+// returns it with any trailing slash stripped. Bare paths are rejected.
+func NormalizePrefix(prefix string) (string, error) {
```
reubeno, line 120: Note: if we end up parsing into a proper URL we should just be able to look at the scheme property, I'd expect.

internal/repo/repolayout/layout.go
```diff
@@ -0,0 +1,162 @@
...
+		return "", errors.New("empty prefix")
+	}
+
+	if !strings.HasPrefix(prefix, "http://") &&
+		!strings.HasPrefix(prefix, "https://") &&
+		!strings.HasPrefix(prefix, "file://") {
+		return "", fmt.Errorf("prefix %#q must be an http://, https://, or file:// URL", prefix)
+	}
+
+	return strings.TrimRight(prefix, "/"), nil
```
reubeno: (Low priority) Do we need to trim the trailing slash?

internal/projectconfig/distro.go
```diff
@@ -86,12 +86,20 @@ type DistroVersionDefinition struct {
...
+	// ordered lists of input references. Each entry references either a
+	// [RpmRepoResource] or a [RpmRepoSet]; sets are expanded at validation time.
 	Inputs DistroVersionInputs `toml:"inputs,omitempty" json:"inputs,omitempty" jsonschema:"title=Inputs,description=Per-use-case input repositories"`
 }
 
+// Use-case identifiers for [DistroVersionInputs]. These match the TOML keys
+// under `[distros.<d>.versions.<v>.inputs]` and are the canonical names used
+// in error messages and CLI flags (e.g. `azldev repo query --use-case`).
+const (
+	UseCaseRPMBuild   = "rpm-build"
```
reubeno, line 99: Thanks for adding the enums :smile:.

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,683 @@
...
+	}
+
+	if len(repos) == 0 {
+		return errors.New("no sub-repos remain after applying filters")
+	}
+
+	workerEnv, cancel := env.WithCancel()
+	defer cancel()
+
+	results := probeAll(workerEnv, env.IOBoundConcurrency(), repos)
```
reubeno: Why do we need to probe the repos? `dnf` will end up doing the same anyhow. Let's see if we can simplify the code that we need to maintain here.

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,683 @@
...
+
+	return cmd
+}
+
+// RunQuery is the entry point for `azldev repo query`. It resolves the
+// template, probes the per-slot repodata URLs, then execs `dnf` with the
+// surviving slots wired up via --repofrompath / --enablerepo. It does not
+// return on success — control is transferred to dnf via [syscall.Exec].
+func RunQuery(env *azldev.Env, options *QueryOptions, dnfArgs []string) error {
+	if !env.CommandInSearchPath(DnfBinary) {
```
reubeno: There's a prereqs package in `azldev` that we should try to use here.

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,683 @@
...
+
+	argv := buildDNFArgv(kept, dnfArgs)
+
+	dnfPath, err := exec.LookPath(DnfBinary)
+	if err != nil {
+		return fmt.Errorf("failed to locate %s in PATH: %w", DnfBinary, err)
+	}
+
+	// Hand control to dnf — does not return on success.
+	if err := syscall.Exec(dnfPath, argv, os.Environ()); err != nil {
```
reubeno: Why `exec`? Can we use the `externalcmd` package and retain control after it exits?

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,683 @@
...
+			strings.Join(failures, "\n  "))
+	}
+
+	if len(kept) == 0 {
+		return errors.New("no reachable sub-repos")
+	}
+
+	argv := buildDNFArgv(kept, dnfArgs)
+
+	dnfPath, err := exec.LookPath(DnfBinary)
```
reubeno: Is this required? Shouldn't execution do the right searching through paths.

internal/app/azldev/cmds/repo/query.go
```diff
@@ -0,0 +1,683 @@
...
+	useCase := options.UseCase
+	if useCase != projectconfig.UseCaseRPMBuild && useCase != projectconfig.UseCaseImageBuild {
+		return nil, fmt.Errorf("--use-case %#q is invalid (want %q or %q)",
+			useCase, projectconfig.UseCaseRPMBuild, projectconfig.UseCaseImageBuild)
+	}
+
+	cfg := env.Config()
+	version := options.Version
+
+	distroName := cfg.Project.DefaultDistro.Name
```
reubeno, line 258: Could you file a follow-up GitHub issue for us to allow overriding the "default distro" on the azldev command line. Right now I think it's only configurable via TOML files.

internal/repo/repolayout/layout.go
liunan-ms: Added `defaultconfigs.DefaultRpmRepoSetTemplateName` in defaultconfigs.go. All references in query.go and layout_test.go are updated.

internal/repo/repolayout/layout.go
liunan-ms: Dropped PrefixIndex/PrefixCount, stamp the dnf RepoID at construction time in buildInputRepos so downstream stages don't need to know prefix order.

internal/repo/repolayout/layout.go
liunan-ms: Fixed

internal/repo/repolayout/layout.go
liunan-ms: Replaced with lo.UniqBy()

internal/repo/repolayout/layout.go
liunan-ms: Done. `NormalizePrefix` now does `url.Parse(prefix)` and validates the scheme against {"http","https","file"}.

internal/repo/repolayout/layout.go
liunan-ms: The explicit `strings.TrimRight` is removed and `url.JoinPath` handles separator normalization. Tests are updated as well.

internal/app/azldev/cmds/repo/query.go
liunan-ms: Removed probing code, `buildDNFArgv` now emits `--setopt=<id>.skip_if_unavailable=1` per repo so dnf logs and skips ones that fail to load.

internal/app/azldev/cmds/repo/query.go
liunan-ms: Fixed

internal/app/azldev/cmds/repo/query.go
liunan-ms: Replaced the `syscall.Exec` block with `env.Command(...).Run(env)`. The factory is backed by externalcmd.

internal/app/azldev/cmds/repo/query.go
liunan-ms: Removed `exec.LookPath(DnfBinary)` together with the above fix using prereqs.

internal/app/azldev/cmds/repo/query.go
liunan-ms: https://github.com/microsoft/azure-linux-dev-tools/issues/226

Issue #203, opened 2026-05-21
`spec-remove-subpackage` drops `%define` macros inside the removed subpackage that are referenced elsewhere


PawelWMS:
## Summary

`spec-remove-subpackage` deletes the entire `%package` / `%description` / `%files` block of the targeted subpackage but does not preserve `%define` (or `%global`) macros that live inside that block when they are referenced from sections that survive removal (`%install`, `%build`, `%check`, etc.). The macro is silently dropped, the referencing section keeps the now-undefined `%{macro}` references, and the build fails far away -- for instance with `Installed (but unpackaged) files found:` on the payload that the surviving section was supposed to clean up.

## Repro

Component: `qemu` on the `4.0` branch of `microsoft/azurelinux`.

Upstream `qemu.spec` (Fedora rawhide, commit `f2cb21ff470d94a6dfd2a2e6ef9effeaa3aead1f`) defines `%{testsdir}` inside the `%package tests` block:

```spec
%package tests
Summary: tests for the %{name} package
Requires: %{name} = %{evr}

%define testsdir %{_libdir}/%{name}/tests-src

%description tests
The %{name}-tests rpm contains tests ...
```

`%{testsdir}` is then referenced from `%install`, which is not guarded by any "tests enabled" conditional — the upstream `%install` unconditionally creates `%{buildroot}%{testsdir}/...` and copies the test fixtures there:

```spec
%install
...
mkdir -p %{buildroot}%{testsdir}/python
mkdir -p %{buildroot}%{testsdir}/tests
mkdir -p %{buildroot}%{testsdir}/tests/qemu-iotests
mkdir -p %{buildroot}%{testsdir}/scripts/qmp

cp -R %{qemu_kvm_build}/python/qemu %{buildroot}%{testsdir}/python
cp -R %{qemu_kvm_build}/scripts/qmp/* %{buildroot}%{testsdir}/scripts/qmp
install -p -m 0755 tests/Makefile.include %{buildroot}%{testsdir}/tests/
...
```

Apply this overlay:

```toml
[[components.qemu.overlays]]
description = "Remove qemu-tests subpackage"
type = "spec-remove-subpackage"
package = "tests"
```

Render with `azldev comp render -p qemu` and inspect `specs/q/qemu/qemu.spec`.

`azldev` version reproduced against:

```
azldev version v0.1.1-0.20260521162032-fc8fadad257d
```

### Expected

[...]

Issue #193, opened 2026-05-14
spec-remove-subpackage / spec-remove-section greedily consume %if guards of the following subpackage


PawelWMS:
## Summary

`spec-remove-subpackage` (and `spec-remove-section` for `%description` /
`%files`) deletes everything from the matched section directive up to the
next section directive (`%package`, `%description`, `%files`, `%install`,
`%build`, etc.). The engine does not understand `%if` / `%else` / `%endif`
as structural elements, so any `%if %{...}` directive sitting between the
removed section and the next subpackage's directive is silently swallowed,
leaving the matching `%endif` as an orphan further down the spec.

## Reproducer

Component: `qemu` on the `tomls/base/main` branch of `microsoft/azurelinux`.
Upstream `qemu.spec` layout around the `tests` subpackage (slightly
simplified):

```spec
%package tests
Summary: tests for the %{name} package
Requires: %{name} = %{evr}

%define testsdir %{_libdir}/%{name}/tests-src

%description tests
The %{name}-tests rpm contains tests ...

%if %{have_libblkio}
%package  block-blkio
Summary: QEMU blkio block driver
...
%endif
```

Apply this overlay:

```toml
[[components.qemu.overlays]]
description = "Drop qemu-tests subpackage"
type = "spec-remove-subpackage"
package = "tests"
```

Render with `azldev comp render -p qemu`. Inspect `specs/q/qemu/qemu.spec`
and you'll see:

1. `%package tests`, `%description tests`, `%files tests` are gone — good.
2. The `%if %{have_libblkio}` lines that originally guarded
   `%package block-blkio` and `%files block-blkio` are ALSO gone — bad.
3. The matching `%endif`s further down (after the `%description block-blkio`
   body and after `%files block-blkio`) are now orphans.
4. `rpmspec -P specs/q/qemu/qemu.spec` fails with `%endif with no %if`.

Same behaviour happens if you use three `spec-remove-section` overlays
(`%package`/`%description`/`%files` each scoped with `package = "tests"`).

## Why this matters

The pattern (an `%if %{have_FOO}` guard right before a subpackage's
`%package`, paired with an `%endif` at the end of that subpackage) is
extremely common in Fedora dist-git specs. Any AZL component that wants to
remove a subpackage that's adjacent (in spec ordering) to a bcond'd
subpackage hits this. The workaround is to add `spec-append-lines` overlays
that re-insert the `%if %{have_FOO}` line into the now-trailing
[...]

PR #215, opened 2026-05-30
feat: add structural tree parser for conditional-aware spec editing

### LINKED ISSUE #214: Structural tree parser for conditional-aware spec editing
## Problem

The spec overlay system operates on section ranges determined by `Spec.Visit`, a single-pass line walker that treats `%if/%endif` conditionals as opaque content. This causes a family of bugs at section boundaries involving conditionals:

| # | Problem | Impact |
|---|---------|--------|
| 1 | `spec-append-lines` places lines inside straddling conditionals | Appended content becomes conditional when it shouldn't be |
| 2 | `collectSectionRanges` needs post-hoc `balanceRange` | Complexity / fragility |
| 3 | `spec-insert-tag` needs `skipPastConditional` workaround | Complexity / fragility |
| 4 | `%define`/`%global` inside removed sections silently dropped | Build failures downstream (#203) |
| 5 | `%else` branches with section headers create phantom sections | Potential incorrect operations |
| 6 | Continuation lines (`\`) misinterpreted as structural elements | Phantom section boundaries, false-positive errors |
| 7 | Section content spilling past wrapper `%endif` invisible to section-scoped operations | Section-scoped overlays miss post-wrapper content |

These patterns are common in Fedora specs (13 specs with `%else` branches containing sections, 22+ specs with continuation-line issues, 529 specs with macro-generated sections). As Azure Linux imports more complex packages, these issues surface increasingly.

**Related issues:** #144, #193, #203

## Proposed solution

Replace the flat Visit-based line walker with a **structural tree parser** (Option B from [RFC 001](https://github.com/trungams/azldev/blob/tvuong/feat/spec-visit/docs/developer/rfc/001-conditional-aware-spec-editing.md)). The tree model makes conditional-boundary bugs structurally impossible rather than patching around them.

### How it works

A two-pass parser builds a `block` tree from raw spec lines:

- **Pass 1**: Collects `%if`/`%endif` pairs and section header positions, skipping `%if`/`%endif` inside `%define`/`%global` macro continuation bodies (those are macro text, not structural conditionals).
- **Pass 2**: Builds nested block tree, classifying conditionals as **wrappers** (contain section headers) vs **content blocks** (fully within a section).

Operations work on the tree via `specTree` / `sectionHandle` types, then serialize back to lines. The tree is an internal implementation detail — the public `Spec.*` API is unchanged.

### What it fixes

- **Boundary bugs (1-3)**: Structurally impossible — sections and conditionals are separate tree nodes, so `AppendLinesToSection` can't land inside a wrapper.
- **Macro hoisting (4)**: `RemoveSections` walks removed subtrees for `%define`/`%global` definitions, scans survivors for references, and hoists referenced macros to the root level automatically.
- **Continuation awareness (6)**: `collectConditionalPairs` skips `%if`/`%endif` inside `%define`/`%global` continuation bodies. `findSectionHeaderLines` skips all continuation lines.
- **`SearchAndReplace` coverage**: A dedicated `searchReplaceBlock` walker visits ALL line types (text, macros, conditional headers, `%endif`) so patterns targeting `%define`/`%global` or `%if` lines work correctly.
- **Visit removal**: `Visit`, `VisitTags`, and `VisitTagsPackage` are replaced by tree-based operations + a new `GetTag` read-only accessor. The entire Visit infrastructure (`parseState`, `Context`, `VisitTarget`, etc.) is deleted.

### Validation

Full render against the Azure Linux 4.0 corpus (~7,400 components):

| Metric | Baseline | Tree branch | Delta |
|--------|----------|-------------|-------|
| Wall time | 24m 11s | 22m 59s | -5.0% |
| CPU (user) | 303m 34s | 277m 07s | -8.7% |
| Render failures | 0 | 0 | — |

Branch [`tvuong/4.0/feat/azldev-parsetree`](https://github.com/microsoft/azurelinux/tree/tvuong/4.0/feat/azldev-parsetree) shows the rendered spec output with the tree-based azldev.

### Known limitations

Documented in the [overlays user guide](https://github.com/trungams/azldev/blob/tvuong/feat/spec-visit/docs/user/reference/config/overlays.md#known-limitations):

1. **Straddling sections**: When a section header is inside a `%if` wrapper but its content continues past `%endif`, section-scoped overlays (`spec-remove-tag`, etc.) cannot find post-wrapper content. Workaround: use `spec-search-replace` with anchored regex. Affects 1 spec (gdb).

   ```spec
   %if 0%{!?scl:1}
   %package headless
   Requires: binutils
   %endif
   # Still %package headless in RPM's view, but the tree parser
   # can't associate this with "headless" — the section header
   # is structurally inside the wrapper.
   Recommends: default-yama-scope
   ```

   The tree parser cannot determine which section owns post-`%endif` content without evaluating the `%if` condition — the same content could belong to `%package headless` (when true) or to whatever section precedes the wrapper (when false).

2. **Macro-generated sections**: Sections created by macros like `%ghc_lib_subpackage` or `%fontpkg` are invisible to the static parser. Use `spec-search-replace` for these.

### Test coverage

- **13 curated fixtures** distilling real-world Fedora/Azure Linux patterns (continuation macros, elif chains, straddling wrappers, nested wrappers, script-section tag-shaped lines, subpackage macro hoisting)
- **Tier 1**: Lossless round-trip (byte-for-byte parse/serialize identity) over every fixture
- **Tier 3**: Per-pattern edit-operation regression tests (tag insertion, HasSection through wrappers, append through conditionals, script-section safety, section-scoped SearchAndReplace)
- **Synthetic stress**: 64 deterministic round-trip seeds + 32 AddTag seeds composing primitives randomly
- **Macro hoisting**: Positive (referenced macro hoisted) and negative (unreferenced macro dropped) regression tests

trungams:
## Summary

Replaces the flat Visit-based spec line walker with a structural tree parser for all spec editing operations. This makes conditional-boundary bugs structurally impossible and fixes several long-standing issues (#144, #193, #203).

Closes #214 

### New: tree parser (`tree.go`)

Two-pass parser that builds a `block` tree from raw spec lines:
- Pass 1: collects `%if`/`%endif` pairs, skipping pairs inside `%define`/`%global` macro continuation bodies
- Pass 2: builds nested block tree, classifying conditionals as wrappers (contain section headers) vs content blocks

Block kinds: `rootBlock`, `sectionBlock`, `conditionalBlock`, `textBlock`, `macroDefBlock`. Lossless round-trip via `serializeTree`.

### New: tree API (`tree_api.go`)

`specTree` / `sectionHandle` types wrap the parsed tree. `Spec.mutateTree` handles parse → mutate → serialize; `Spec.inspectTree` for read-only access. The public `Spec.*` API is unchanged for callers.

### New: `SearchAndReplace` rewrite (`searchReplaceBlock`)

Dedicated recursive walker that visits ALL line types (text, macros, conditional headers, `%endif`). The old `VisitAllLines`-based implementation skipped `%define`/`%global` and conditional directive lines. Also relaxes the section filter for wrapper else-branch content.

### New: macro hoisting (`tree_hoist.go`)

`RemoveSections` now automatically hoists `%define`/`%global` macros from removed sections when they're referenced by surviving content (#203). Scans for `%{name}`, `%{?name}`, `%{!?name}`, bare `%name`, and conditional headers like `%if 0%{?with_name}`.

### New: `GetTag` and `PrependLinesToAllSections`

- `GetTag(packageName, tag)` — read-only tag lookup via `inspectTree`, replaces `VisitTagsPackage` usage
- `PrependLinesToAllSections` + `spec-prepend-all-lines` overlay — prepends to every matching section (used by check-skip to disable all `%check` sections)

### New: testdata fixtures + synthetic stress

13 curated `.spec` fixtures distilling real-world patterns. Two test tiers (lossless round-trip + per-pattern edit-operation regression). Synthetic generator with 64+32 deterministic seeds.

### Removed: Visit infrastructure

`Visit`, `VisitTags`, `VisitTagsPackage`, `Context`, `VisitTarget`, `VisitTargetType`, `Visitor`, `parseState`, `parseSpecLine`, `newParsedLine`, `parseLogicalLine` — all deleted. The sole external caller (`release.go`) migrated to `GetTag`.

Also removed: `sectionLineRange`, `collectSectionRanges`, `balanceRange`, `validateNoContentAfter`, `skipPastConditional`, `findInsertTagPosition`, `isBlankOrComment`, `removeRanges`, and related helpers.

### Docs

- RFC 001 added to `docs/developer/rfc/`
- Known limitations documented in `docs/user/reference/config/overlays.md`

## Validation

Full render against Azure Linux 4.0 (~7,400 components):

| Metric | Baseline | This PR | Delta |
|--------|----------|---------|-------|
| Wall time | 24m 11s | 22m 59s | -5.0% |
| CPU (user) | 303m 34s | 277m 07s | -8.7% |
| Render failures | 0 | 0 | — |

Rendered output: [`tvuong/4.0/feat/azldev-parsetree`](https://github.com/microsoft/azurelinux/tree/tvuong/4.0/feat/azldev-parsetree)

## Known limitations

1. **Straddling sections** — section header inside `%if` wrapper with content past `%endif` is invisible to section-scoped overlays. Workaround: `spec-search-replace`. Affects 1 spec (gdb).
2. **Macro-generated sections** — sections from `%ghc_lib_subpackage`, `%fontpkg`, etc. are invisible to the static parser.


trungams:
The gdb render failure in e2e is expected as we'll need to update the overlay

docs/developer/rfc/001-conditional-aware-spec-editing.md
```diff
@@ -0,0 +1,559 @@
...
+### Non-issues: patterns already handled correctly
+
+| Pattern | Prevalence | Why it's fine |
+|---------|-----------|---------------|
+| `%dnl` (comment directive) | 5 specs | Parser sees `%dnl` as first token → not in `sectionTypesByName` → raw line |
+| `%bcond_with`/`%bcond_without` | Pervasive | Purely macro-level; structural impact is through `%if %{with ...}` which is already handled |
+| `%{lua:...}` blocks | Pervasive (via `%autorelease`) | Inline macro expansion; parser sees the containing `%define` line |
+| Comments (`#` lines) inside conditionals | Pervasive | Parser treats as empty lines; `conditionalDepthChange` ignores them |
+
+### Summary
```
dmcilvaney: Could we add some mechanism to control which branch we go down? For example, many specs have something like
```spec
%if <condition>
Release: 1
%else
Release: %autorelease
%endif
```
It would be nice to be able to target one.
Maybe a mode enum? Something like `First`, `All`, `SpecificN`, `RequireUnique` maybe?

This could apply generally to both read and write operations.

trungams: It's definitely possible since each branch will be its own block. My question is whether we want to allow users to select branches based on condition (requires evaluation) or just ordering of the branches. Also how should we target these conditional blocks, since there may be multiple with similar conditions in the same spec file.

docs/developer/rfc/001-conditional-aware-spec-editing.md
```diff
@@ -0,0 +1,559 @@
...
+
+### What remains broken
+
+1. **`spec-append-lines` boundary bug**: `AppendLinesToSection` uses `SectionEndTarget` to find where to insert. When a `%if` sits between two sections (wrapping the next one), Visit considers it part of the current section, so appended lines land inside the conditional. `spec-insert-tag` has a workaround (`skipPastConditional`); `spec-append-lines` does not.
+
+2. **`%define`/`%global` dropped during removal** (#203): `spec-remove-subpackage` deletes the entire block without checking whether macro definitions inside it are referenced from surviving sections. The macro silently vanishes, causing build failures downstream.
```
dmcilvaney: Probably out of scope for this doc, but curious how we might handle recursive definitions

```spec
%define a %{b}
%define b some_value
```

If I understand the algorithm right we would move `%define a %{b}` to the global scope, but might now have a dangling link to `%{b}`.

docs/developer/rfc/001-conditional-aware-spec-editing.md
dmcilvaney: Is there a way to map the blocks back to the line numbers? Maybe have each block have a line offset inside its parent? That would provide a very basic way to find `First` and `SpecificN` (if we even WANT that one) based on sorting. `All` just applies everywhere (I think this is sort of the implicit default right now), and `RequireUnique` just errors out if there is any ambiguity.

docs/developer/rfc/001-conditional-aware-spec-editing.md
```diff
@@ -0,0 +1,559 @@
...
+
```
dmcilvaney: Also, should hoisting be an opt-in flag, or on by default? Direction I've been getting is "no magic" because it gets confusing to reason about. Randomly moving variables might be surprising. If the flag is instead a list, would also fix the recursive definition thing above.

internal/rpm/spec/tree_hoist.go
```diff
@@ -0,0 +1,239 @@
...
+func blockSet(blocks []*block) map[*block]bool {
+	set := make(map[*block]bool, len(blocks))
+	for _, b := range blocks {
+		set[b] = true
+	}
+
+	return set
+}
+
+// collectMacrosInSections gathers every [macroDefBlock] reachable from any of
```
dmcilvaney, line 356: This pattern seems to repeat a bunch, thoughts on making a `walkBlocks()` function, or at least an interface.

```go
func collectMacrosInSections(sections []*block) []*block {
    var macros []*block
    for _, sec := range sections {
        walk(sec, func(b *block) bool {
            if b.Kind == macroDefBlock {
                macros = append(macros, b)
            }
            return true
        })
    }
    return macros
}
```

Mutating funcs (removeBlocKFromParent etc.) feel harder, maybe leave them as-is?

docs/developer/rfc/001-conditional-aware-spec-editing.md
trungams: I addressed this in the latest set of changes. We should keep track of these definitions and hoist them altogether if there are dependencies that need to stay

docs/developer/rfc/001-conditional-aware-spec-editing.md
trungams: I think it should be on by default. I've added some logs whenever we move some macro defs

internal/rpm/spec/tree_hoist.go
trungams: Updated

docs/developer/rfc/001-conditional-aware-spec-editing.md
trungams: hmm, the blocks are already sorted via in-order traversal, so we don't need exact line number matching/offset to do `First`/`SpecificN` stuff. Once we figure out which overlay operation we want to support with this mechanism I can always add it in a follow up PR

internal/rpm/spec/tree.go
```diff
@@ -0,0 +1,852 @@
...
+		condBlock.Else = elseContainer.Children
+	}
+
+	return nil
+}
+
+// isElifDirective returns true if the line is a %elif/%elifarch/%elifnarch/%elifos/%elifnos
+// directive (as opposed to a plain %else which is a terminal branch).
+func isElifDirective(rawLine string) bool {
+	lower := strings.ToLower(strings.Fields(strings.TrimSpace(rawLine))[0])
```
liunan-ms: A length check can be added to prevent strings.Fields(...)[0] panic on a whitespace-only line.

internal/rpm/spec/tree_hoist.go
```diff
@@ -0,0 +1,369 @@
...
+	return ordered
+}
+
+// hoistIntoPreamble appends the given macro blocks to the end of the preamble
+// section (the implicit section before the first section header), so they are
+// defined before any section that might reference them. If no preamble section
+// exists, the macros are prepended at the root as a fallback.
+func hoistIntoPreamble(root *block, macros []*block) {
+	if preamble := findSectionBlock(root, "", ""); preamble != nil {
+		preamble.Children = append(preamble.Children, macros...)
```
liunan-ms, line 321: Is the following case where `bar` is undefined at the point `foo` is expanded possible? is this an acceptable case or a limitation?
```
%global foo %{bar}
...
%package tests
%global bar value 
```

liunan-ms:
Thank you Trung for adding the RFC and the structural tree parsing! I like the design, just left two comments/questions. NIT(non-blocking): another thing we can do is splitting the large refactor PR into 3 PRs: 
one for adding `collectConditionalPairs` macro-continuation fix (as RFC mentioned the continuation/inContinuation fix is independent of the structural model work and could ship as a standalone PR), 
one for adding structural tree parser with logical commits (still large I know but they are the complete refactor), 
then a separate PR for adding `spec-prepend-all-lines` overlay.

PR #216, opened 2026-06-01
fix: disable validation checks when passing --permissive-config

dmcilvaney:
Needed for some future work to load and apply historic overlays.

### DIFF
--- a/internal/projectconfig/project.go
+++ b/internal/projectconfig/project.go
@@ -6,7 +6,6 @@ package projectconfig
 import (
 	"errors"
 	"fmt"
-	"log/slog"
 	"sort"
 
 	"dario.cat/mergo"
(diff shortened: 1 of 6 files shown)


internal/projectconfig/project.go
```diff
@@ -68,23 +69,43 @@ func NewProjectConfig() ProjectConfig {
...
+// Validate performs semantic validation of the configuration, returning an error if any
+// semantic errors are found. When permissive is true, cross-reference consistency checks
+// (component-group membership, package-group membership, and image/test-suite references)
+// are downgraded to informational logs rather than hard errors.
+//
+// Permissive validation exists for best-effort loads such as historical overlay replay,
+// where a partial or point-in-time config may legitimately reference entities that are
+// defined in a different revision. Structural validation (required fields, value formats)
+// is always enforced.
+func (cfg *ProjectConfig) Validate(permissive bool) error {
```
reubeno: I think it would be cleaner (and less code churn) to have the caller ignore errors in permissive mode. What do you think?

reubeno (changes requested):

internal/projectconfig/project.go
dmcilvaney: Sure, I can do that

reubeno (approved):

PR #217, opened 2026-06-01
feat: add historic git-backed filesystem

dmcilvaney:
When creating synthetic dist-gits, there are certain packages that use overlays which mutate the calculated release value for a package. If the overlays are only applied to the last commit in the dist-git, then if a package changes the 'version' of a package the autorelease tool will see the version change ONLY on that last commit, and conclude the release should be reset back to '1'. If we then add another commit to the dist-git, the version should be '2', but the autorelease tool will still see the version bump happen only on the last commit, and again reset the release to '1'. (see kernel-headers for an example)

To avoid this, we must apply the overlays to all commits in the dist-git, as they would have appeared at the time of the commit. azldev already abstracts filesystem access through 'fs', so we can implement a 'gitfs' that loads toml configs from the project's git history. A copy-on-write in-memory filesystem allows for writing if necessary.

### DIFF
--- a/internal/projectconfig/historic.go
+++ b/internal/projectconfig/historic.go
@@ -93,6 +93,13 @@ func LoadProjectConfigAtCommit(
 // the historic tree), and distro defaults are not used for version-setting
 // overlays. This keeps historic resolution self-contained and deterministic.
 //
+// Each call performs a full LoadProjectConfigAtCommit (fresh overlay, re-staged
+// defaults, re-parsed config) to extract a single component, so resolving many
+// components at one commit reloads the project repeatedly. This favors a simple,
+// self-contained API over performance; the currently expected workflows resolve
+// few components per commit. If a caller needs many-per-commit resolution, load
+// the config once and resolve against the returned *ProjectConfig instead.
+//
 // Returns (nil, nil) when the component is absent at that commit.
 func ResolveComponentOverlaysAtCommit(
 	repo *gogit.Repository,
--- a/internal/projectconfig/historic.go
+++ b/internal/projectconfig/historic.go
@@ -1,0 +1,117 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package projectconfig
+
+import (
+	"fmt"
+
+	gogit "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/gitfs"
+	"github.com/spf13/afero"
+)
+
+// historicDryRunnable reports that we are not in dry-run mode: the historic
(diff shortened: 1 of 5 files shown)


dmcilvaney:
Requires https://github.com/microsoft/azure-linux-dev-tools/pull/216 for tests to pass

internal/utils/gitfs/gitfs.go
```diff
@@ -0,0 +1,482 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Package gitfs provides a read-only [afero.Fs] backed by a git tree at a
```
reubeno, line 4: Did we look to see if any of the other OSS afero.FS implementations already do this?

internal/utils/gitfs/gitfs.go
```diff
@@ -0,0 +1,482 @@
...
+	default:
+		return fileperms.ReadOnlyFile
+	}
+}
+
+//
+// Mutating operations — all unsupported.
+//
+
+func (f *Fs) Create(name string) (afero.File, error) {
```
reubeno, line 244: Should we consider looking at `afero.NewReadOnlyFS()`? Understood that we still need these stubs.

internal/projectconfig/historic.go
```diff
@@ -0,0 +1,134 @@
...
+//
+// Each call performs a full LoadProjectConfigAtCommit (fresh overlay, re-staged
+// defaults, re-parsed config) to extract a single component, so resolving many
+// components at one commit reloads the project repeatedly. This favors a simple,
+// self-contained API over performance; the currently expected workflows resolve
+// few components per commit. If a caller needs many-per-commit resolution, load
+// the config once and resolve against the returned *ProjectConfig instead.
+//
+// Returns (nil, nil) when the component is absent at that commit.
+func ResolveComponentOverlaysAtCommit(
```
reubeno, line 104: Why so specific to only return overlays? It's already doing the work to get the full component info and then discarding it -- could it just return the component config and let the caller extract the overlays if it wants?

internal/projectconfig/historic.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (historicOSEnv) Getwd() (string, error) { return "", nil }
+func (historicOSEnv) Chdir(string) error     { return nil }
+func (historicOSEnv) Getenv(string) string   { return "" }
+func (historicOSEnv) IsCurrentUserMemberOf(string) (bool, error) {
+	return false, nil
+}
+func (historicOSEnv) LookupGroupID(string) (int, error) { return 0, nil }
+
+// LoadProjectConfigAtCommit loads the project configuration exactly as it
```
reubeno, line 37: Will this require us to be backwards-compatible forever? That's a higher bar than we otherwise have right now -- and one that I'm still not sure we're ready to take on strongly.

PR #218, opened 2026-06-02
perf(comp list): skip lock file reads (non-mutating)

binujp:
I hit a roadblock with `azldev comp list -a` performance while working on some tooling. This was on wsl with workspace on a windows mount. This was due to slow fs syscall performance. On average about 40 seconds for this to complete. The same would complete under 2s on a linux native filesystem. Since list and other similar operations which do not need to populate lock [meta]data a simple bypass results in a large performance gain. Though the fix is for wsl environment the benefit is seen in other cases too with under 200ms completion.

```
+---------+------------------+------+-----+------+------+
|         |                  |   WSL      |   Native    |
|         |                  +------+-----+------+------+
|         |                  | COLD | HOT | COLD | HOT  |
+---------+------------------+------+-----+------+------+
| current | azldev           |    7 |   3 | 0.39 | 0.15 |
| current | azldev comp list | 100+ |  35 | 3.4  | 0.38 |
| new     | azldev           |    3 |   3 | 0.39 | 0.15 |
| new     | azldev comp list |  3.2 | 3.2 | 0.29 | 0.27 |
+---------+------------------+------+-----+------+------+
```
Add SkipLockPopulation flag to the component Resolver that bypasses populateFromLock() during component resolution. This eliminates thousands of unnecessary lock file reads (one per component) for commands that never consume the lock data.

Commands updated:
- 'component list': read-only, returns configs without lock state
- 'component changed': reads locks directly from git commits, not disk

Performance impact (measured on Azure Linux repo with ~7k components):
- 'component list -a' reduced from ~38s to ~8s (lock reads were the dominant I/O cost at ~30s)

### DIFF
--- a/internal/app/azldev/cmds/component/changed.go
+++ b/internal/app/azldev/cmds/component/changed.go
@@ -119,11 +119,13 @@ const (
 func ChangedComponents(
 	env *azldev.Env, options *ChangedComponentOptions,
 ) ([]ChangedResult, error) {
-	// Changed compares lock files between git refs — skip validation since
-	// the current working-tree locks may legitimately be stale.
-	options.ComponentFilter.SkipLockValidation = true
+	// Changed compares lock files between git refs — skip both validation and population
+	// since it reads locks directly from git commits (ReadAllAtCommit).
+	options.ComponentFilter.LockMode = components.LockModeSkipBoth
 
 	resolver := components.NewResolver(env)
+	// Skip lock population since changed reads locks directly from git commits
+	resolver.SkipLockPopulation = true
 
 	comps, err := resolver.FindComponents(&options.ComponentFilter)
 	if err != nil {
--- a/internal/app/azldev/cmds/component/list.go
+++ b/internal/app/azldev/cmds/component/list.go
@@ -17,6 +17,9 @@ import (
 type ListComponentOptions struct {
 	// Standard filter for selecting components.
 	ComponentFilter components.ComponentFilter
+	// SkipLockPopulation controls whether lock files are read during resolution.
+	// Only applicable when component list is explicitly requested by the user.
+	SkipLockPopulation bool
 }
 
 func listOnAppInit(_ *azldev.App, parentCmd *cobra.Command) {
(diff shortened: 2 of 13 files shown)


internal/app/azldev/cmds/component/list.go
```diff
@@ -77,6 +77,9 @@ func ListComponentConfigs(
 	options.ComponentFilter.SkipLockValidation = true
 
 	resolver := components.NewResolver(env)
+	// List doesn't consume lock data — skip lock population to avoid
+	// reading thousands of lock files unnecessarily.
+	resolver.SkipLockPopulation = true
```
dmcilvaney: Could we get a flag to re-enable this behaviour? (or maybe the flag tuns it off?).
This is the only easy way to get the resolved lock info for a component (via -O json). I've used it a bunch for debugging.

internal/app/azldev/core/components/resolver.go
```diff
@@ -41,6 +41,11 @@ type Resolver struct {
 	// Only 'component update' enables this — it uses the freshness status to
 	// decide which components need re-resolution and which can be skipped.
 	CheckFreshness bool
+	// SkipLockPopulation skips reading lock files and attaching lock data to
+	// resolved components. Set this for read-only commands (list, changed)
+	// that don't consume [projectconfig.ComponentLockData] — it eliminates
+	// potentially thousands of lock file reads.
+	SkipLockPopulation bool
```
dmcilvaney, line 50: There are a bunch of weird combos we can get now with the flags, might be better to use an enum. For example, you can have `SkipLockPopulation=true` and `SkipLockValidation=false` which I think still hits the perf issue, and doesn't really make sense to me.

internal/app/azldev/cmds/component/changed.go
```diff
@@ -124,6 +124,9 @@ func ChangedComponents(
 	options.ComponentFilter.SkipLockValidation = true
 
 	resolver := components.NewResolver(env)
+	// Changed reads locks directly from git commits (ReadAllAtCommit), so
+	// it doesn't need the resolver to populate lock data from disk.
+	resolver.SkipLockPopulation = true
```
dmcilvaney, line 128: We have `parmap`, and in theory everything in the lock store should be thread-safe.
Could you quickly check what the perf is of using that?
```diff
diff --git a/internal/app/azldev/core/components/resolver.go b/internal/app/azldev/core/components/resolver.go
index 41ab031..c752513 100644
--- a/internal/app/azldev/core/components/resolver.go
+++ b/internal/app/azldev/core/components/resolver.go
@@ -4,6 +4,7 @@
 package components
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -16,6 +17,7 @@ import (
 	"github.com/microsoft/azure-linux-dev-tools/internal/fingerprint"
 	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
 	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileutils"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/parmap"
 )
 
 // Well-known errors.
@@ -103,6 +105,8 @@ func (r *Resolver) FindComponents(filter *ComponentFilter) (components *Componen
 		}
 	}
 
+	r.populateLocksParallel(components)
+
 	r.warnOnLockDrift(components)
 
 	return components, r.validateLockFiles(components, false, skipValidation)
@@ -162,6 +166,8 @@ func (r *Resolver) FindAllComponents() (components *ComponentSet, err error) {
 		components.Add(comp)
 	}
 
+	r.populateLocksParallel(components)
+
 	return components, nil
 }
 
@@ -497,17 +503,23 @@ func (r *Resolver) createComponentFromConfig(componentConfig *projectconfig.Comp
 		componentConfig.Autospec.ReleaseCalculation = projectconfig.ReleaseCalculationAuto
 	}
 
-	// Populate locked state onto the component config. This makes lock data
-	// available to all downstream consumers (render, build, prepare-sources,
-	// diff-sources) without each needing lock-file awareness.
-	r.populateFromLock(componentConfig)
-
 	return &resolvedComponent{
 		env:    r.env,
 		config: *componentConfig,
 	}, nil
 }
 
+func (r *Resolver) populateLocksParallel(set *ComponentSet) {
+	comps := set.Components()
+
+	_ = parmap.Map(r.env, r.env.IOBoundConcurrency(), comps, nil,
+		func(_ context.Context, c Component) error {
+			r.populateFromLock(c.GetConfig())
+
+			return nil
+		})
+}
+
 // populateFromLock reads lock file data and attaches it to the component config
 // as [projectconfig.ComponentLockData]. This centralizes lock-file consumption so
 // all downstream commands (render, build, prepare-sources, diff-sources) get
```

binujp: With parallel populate cold list -a (no cache effect) time was around 21s and hot was around 10s. There is definitely an improvement in the cold start time. It is still too much to make a difference in interactive sessions.

I had tried a parallel file read implementation which did worse and conclusion I came to was that the fs iops was so low the resulting pileup was negatively impacting performance gains from the parallelism.

binujp:
Thought about the comments and the use-case. Given the WSL slowdown is environment specific or a known quantity otherwise, I will re-implement the lock read skipping as an opt-in. That helps preserve existing behavior at a bug compatibility level.

PR #224, opened 2026-06-03
feat(sources): add tarball overlays for source archive modification

Tonisal-byte:
Adds tarball overlay support for modifying source archives by applying file overlays during repack.

```toml
# Remove a single file from inside a source archive
[[components.mypackage.overlays]]
type = "file-remove"
archive = "mypackage-1.0.tar.gz"
file = "vendor/**"
description = "Remove all bundled vendor files"

# Search-and-replace inside a file within an archive
[[components.mypackage.overlays]]
type = "file-search-replace"
archive = "mypackage-1.0.tar.gz"
file = "configure.ac"
regex = "AC_CHECK_LIB\\(old_lib"
replacement = "AC_CHECK_LIB(new_lib"
description = "Update library reference in configure script"

```


internal/utils/tarball/tarball.go
```diff
@@ -0,0 +1,372 @@
...
+
+// RepackDeterministic creates a new tar archive from the contents of sourceDir
+// and writes it to archivePath, replacing any existing file.
+//
+// The output is deterministic:
+//   - File ordering is lexicographic (via [filepath.WalkDir]).
+//   - Timestamps are pinned to Unix epoch (1970-01-01 00:00:00 UTC).
+//   - Owner/group IDs and names are zeroed out.
+//   - Gzip output uses best compression with no OS or filename metadata.
+func RepackDeterministic(fs opctx.FS, archivePath, sourceDir string, comp Compression) (err error) {
```
reubeno: Why is it called "repack" if the input is just a directory? Is there a reason we couldn't use this function for something that didn't originally come from an archive?

internal/utils/tarball/tarball.go
```diff
@@ -0,0 +1,372 @@
...
+	file, err := fs.OpenFile(archivePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, fileperms.PublicFile)
```
reubeno: Why PublicFile?

internal/utils/tarball/tarball.go
```diff
@@ -0,0 +1,372 @@
...
+		return CompressionNone, nil
+	default:
+		return CompressionNone, fmt.Errorf("unsupported archive format %#q", filename)
+	}
+}
+
+// Extract reads a tar archive from the filesystem, decompresses it, and extracts
+// all entries to destDir. Supported entry types are regular files, directories,
+// and symlinks. Path traversal entries are rejected.
+func Extract(fs opctx.FS, archivePath, destDir string, comp Compression) (err error) {
```
reubeno: Is there really not already a golang function (or well-supported 3rd party package) that does this?

internal/utils/tarball/tarball.go
```diff
@@ -0,0 +1,372 @@
...
+	})
+	if walkErr != nil {
+		return fmt.Errorf("walking directory for repacking:\n%w", walkErr)
+	}
+
+	return nil
+}
+
+// ResolveExtractRoot determines the root directory of extracted tarball content.
+func ResolveExtractRoot(workDir string) (string, error) {
```
reubeno: Why is this in this package? It doesn't sound like it has anything to do with tar archives themselves.

internal/utils/tarball/tarball.go
```diff
@@ -0,0 +1,372 @@
...
+
+func writeEntryDeterministic(
+	tarWriter *tar.Writer, path, rel string, entry os.DirEntry, epoch time.Time,
+) error {
+	info, err := entry.Info()
+	if err != nil {
+		return fmt.Errorf("stat %#q:\n%w", path, err)
+	}
+
+	if info.Mode()&os.ModeSymlink != 0 {
```
reubeno: Why is this special-cased?

internal/projectconfig/overlay.go
```diff
@@ -17,10 +17,13 @@ import (
...
 type ComponentOverlay struct {
 	// The type of overlay to apply.
-	Type ComponentOverlayType `toml:"type" json:"type" validate:"required" jsonschema:"enum=spec-add-tag,enum=spec-insert-tag,enum=spec-set-tag,enum=spec-update-tag,enum=spec-remove-tag,enum=spec-prepend-lines,enum=spec-append-lines,enum=spec-search-replace,enum=spec-remove-section,enum=spec-remove-subpackage,enum=patch-add,enum=patch-remove,enum=file-prepend-lines,enum=file-search-replace,enum=file-add,enum=file-remove,enum=file-rename,title=Overlay type,description=The type of overlay to apply"`
+	Type ComponentOverlayType `toml:"type" json:"type" validate:"required" jsonschema:"enum=spec-add-tag,enum=spec-insert-tag,enum=spec-set-tag,enum=spec-update-tag,enum=spec-remove-tag,enum=spec-prepend-lines,enum=spec-append-lines,enum=spec-search-replace,enum=spec-remove-section,enum=spec-remove-subpackage,enum=patch-add,enum=patch-remove,enum=file-prepend-lines,enum=file-search-replace,enum=file-add,enum=file-remove,enum=file-rename,enum=tarball-file-remove,enum=tarball-search-replace,enum=tarball-patch,title=Overlay type,description=The type of overlay to apply"`
```
reubeno: For one thing -- I think we need something that's a bit more abstracted from "tarball". There are also source payloads that are .zip files, or some other format. Are these overlays specific because they operate on resources downloaded from lookaside cache, or are they specific because they operate on files that are archives containing files?

internal/projectconfig/overlay.go
```diff
@@ -119,6 +122,14 @@ func (c *ComponentOverlay) ModifiesSpec() bool {
 		c.Type == ComponentOverlayRemovePatch
 }
 
+// ModifiesTarball returns true if the overlay modifies files inside a source tarball.
+// These overlays require a mock chroot for extraction and repacking.
+func (c *ComponentOverlay) ModifiesTarball() bool {
+	return c.Type == ComponentOverlayTarballFileRemove ||
+		c.Type == ComponentOverlayTarballSearchReplace ||
```
reubeno: Do we need the tarball search/replace or tarball patch? What does it even mean to do those?

internal/app/azldev/core/sources/tarballoverlays.go
```diff
@@ -0,0 +1,358 @@
...
+	ctx context.Context,
+	cmdFactory opctx.CmdFactory,
+	fs opctx.FS,
+	sourcesDirPath string,
+	group tarballGroup,
+) error {
+	archivePath := filepath.Join(sourcesDirPath, group.tarball)
+
+	// Create a temporary directory for extraction.
+	workDir, err := os.MkdirTemp("", "tarball-overlay-")
```
reubeno: Why aren't we using azldev-standard facilities for temp dirs?

internal/app/azldev/core/sources/tarballoverlays.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	defer func() {
+		if removeErr := os.RemoveAll(workDir); removeErr != nil {
+			slog.Warn("Failed to clean up tarball work directory", "error", removeErr)
+		}
+	}()
+
+	// Detect compression and extract.
+	compression, err := tarball.DetectCompression(group.tarball)
```
reubeno: Would all callers need to call DetectCompression followed by Extract? Why not provide a composite call?

internal/app/azldev/core/sources/tarballoverlays.go
```diff
@@ -0,0 +1,358 @@
...
+		return fmt.Errorf("detecting compression for %#q:\n%w", group.tarball, err)
+	}
+
+	if err := tarball.Extract(fs, archivePath, workDir, compression); err != nil {
+		return fmt.Errorf("extracting tarball:\n%w", err)
+	}
+
+	// Determine the root of the extracted content. Most source tarballs have
+	// a single top-level directory (e.g., "pkg-1.0/").
+	extractRoot, err := tarball.ResolveExtractRoot(workDir)
```
reubeno: This seems like a heuristic. Would it be better to require it as an input rather than guessing? Or is there a concern that it will change from version to version and be hard to predict?

internal/app/azldev/core/sources/tarballoverlays.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+
+		return tarballApplyPatch(ctx, cmdFactory, fs, extractRoot, overlay.Source, stripLevel)
+
+	default:
+		return fmt.Errorf("unsupported tarball overlay type %#q", overlay.Type)
+	}
+}
+
+// tarballFileRemove removes files matching a glob pattern from the extracted tree.
```
reubeno: Are we supporting double star globs?

internal/app/azldev/core/sources/tarballoverlays.go
```diff
@@ -0,0 +1,358 @@
...
+	if runErr := cmd.Run(ctx); runErr != nil {
+		return fmt.Errorf("patch failed:\n%s\n%w", stderr.String(), runErr)
+	}
+
+	return nil
+}
+
+// globFilesInDir finds files under root matching a glob pattern.
+// Supports doublestar patterns (e.g., "**/*.md").
+func globFilesInDir(root, pattern string) ([]string, error) {
```
reubeno: Do we not already have helpers to do this?

docs/user/reference/config/overlays.md
```diff
@@ -47,21 +47,34 @@ successfully makes a replacement to at least one matching file.
 | `file-remove` | Removes a file | `file` | Glob pattern for files to remove |
 | `file-rename` | Renames a file within the same directory | `file`, `replacement` | Name of file to rename |
 
+### Tarball Overlays
+
+These overlays modify files **inside** source tarballs. The tarball is extracted into a temporary directory, modifications are applied, and the tarball is repacked with the same compression format. Extraction and repacking are handled natively; patch application requires the `patch` command on the host.
+
+> **Note:** Tarball overlays are applied before spec and file overlays, so subsequent overlays see the modified tarball. The `tarball-patch` overlay type requires the `patch` command to be installed on the host.
```
reubeno: Why is this required? Would it not be simpler to handle these along with other overlays?

docs/user/reference/config/overlays.md
```diff
@@ -47,21 +47,34 @@ successfully makes a replacement to at least one matching file.
 | `file-remove` | Removes a file | `file` | Glob pattern for files to remove |
 | `file-rename` | Renames a file within the same directory | `file`, `replacement` | Name of file to rename |
 
+### Tarball Overlays
```
reubeno: Let's not use "tarball" as the term to associate with this. What if there are zip files? Or some other kind of archive?

docs/user/reference/config/overlays.md
```diff
@@ -47,21 +47,34 @@ successfully makes a replacement to at least one matching file.
...
+
+These overlays modify files **inside** source tarballs. The tarball is extracted into a temporary directory, modifications are applied, and the tarball is repacked with the same compression format. Extraction and repacking are handled natively; patch application requires the `patch` command on the host.
+
+> **Note:** Tarball overlays are applied before spec and file overlays, so subsequent overlays see the modified tarball. The `tarball-patch` overlay type requires the `patch` command to be installed on the host.
+
+| Type | Description | Required Fields |
+|------|-------------|-----------------|
+| `tarball-file-remove` | Removes file(s) matching a glob pattern from inside a tarball | `tarball`, `file` |
```
reubeno: Did we consider alternatively representing this as a plain file-remove overlay, but with an additional option to operate against the contents of a given archive? That would allow reuse of all the existing overlays if we refactor the logic appropriately.

What are the trade-offs between separate overlays vs. reuse existing ones?

docs/user/reference/config/overlays.md
```diff
@@ -47,20 +47,44 @@ successfully makes a replacement to at least one matching file.
 | `file-remove` | Removes a file | `file` | Glob pattern for files to remove |
 | `file-rename` | Renames a file within the same directory | `file`, `replacement` | Name of file to rename |
 
+> **Tip:** `file-remove` and `file-search-replace` can also operate inside a source archive by
+> setting the `archive` field — see [Archive Overlays](#archive-overlays).
+
+### Archive Overlays
+
+A `file-remove` or `file-search-replace` overlay can modify files **inside** a source archive
```
PawelWMS, line 54: Does this support removing/editing files in nested archives? I'm thinking about cases where a.tar.gz contains a/internal/archive.tar.gz and you want to remove a.tar.gz/a/internal/archive.tar.gz/nested/dir/file-to-remove.txt.

If it doesn't, I don't think it's a P0 (I believe all 10 scripts can be replaced without the nested archives supported), but with the new archive field required for differentiating between "regular" files and archives, adding this extension in the future may be tricky.

Considering that, do you think it would be possible to use the file path alone to denote the file to be removed/edited and automatically detect, if a file on the path is a directory vs an archive as in the a.tar.gz/a/internal/archive.tar.gz/nested/dir/file-to-remove.txt example? Starting with this would give us support for nested archives in one go.

PS Cool idea with re-using existing overlays.

docs/user/reference/config/overlays.md
```diff
@@ -274,6 +298,37 @@ description = "Remove CVE patches that are now upstream"
...
+Set the `archive` field on a `file-remove` overlay to delete files matching a glob pattern from
+inside a source archive. The archive is extracted, matching files are removed, and the archive is
+repacked.
+
+```toml
+[[components.mypackage.overlays]]
+type = "file-remove"
+archive = "mypackage-1.0.tar.gz"
+file = "vendor/**"
+description = "Remove bundled vendor directory"
```
PawelWMS: Is this example correct? I'd read `vendor/**` as "remove everything inside the `vendor` directory **BUT** leave the directory there". If I wanted to remove the directory itself, I think I'd go for `vendor`.

Also, in this case, won't a simpler `vendor/*` (single `*` character) work? The `**` glob seems more useful for cases like `vendor/**/test*.abc`, which would allow to remove files matching `test*.abc` from arbitrary subdirectories of `vendor`.

docs/user/reference/config/overlays.md
Tonisal-byte: Thanks! The idea was originally Reubens but I'm glad the implementation is sound. I think it would be better if we revisited this nested archive format in the future. I do agree it's important, but it might introduce complexity to the overlay application process

internal/projectconfig/overlay.go
```diff
@@ -21,6 +21,17 @@ type ComponentOverlay struct {
...
+	// Must be a filename (not a path) matching a source archive in the component's sources directory.
+	// Only file-remove supports archive scoping; when set, it removes file(s) from inside the named
+	// archive instead of the loose sources tree.
+	Archive string `toml:"archive,omitempty" json:"archive,omitempty" jsonschema:"title=Archive,description=The source archive to modify (e.g. pkg-1.0.tar.gz)"`
+	// For overlays that target files inside a source archive, optionally overrides the top-level
+	// directory to treat as the extraction root, mirroring rpmbuild's `%setup -n`. When unset, the
+	// root is inferred: if the archive unpacks to a single top-level directory (the conventional
+	// `%{name}-%{version}` layout) that directory is used; otherwise the archive root is used.
+	// Set this when an archive's top-level directory does not follow that convention.
+	ArchiveRoot string `toml:"archive-root,omitempty" json:"archiveRoot,omitempty" jsonschema:"title=Archive root,description=Top-level directory inside the archive to treat as the extraction root (mirrors %setup -n); inferred when unset"`
```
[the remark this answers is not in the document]
Tonisal-byte: Not necessary

internal/projectconfig/overlay.go
```diff
@@ -21,6 +21,17 @@ type ComponentOverlay struct {
 	// Human readable description of overlay; primarily present to document the need for the change.
 	Description string `toml:"description,omitempty" json:"description,omitempty" jsonschema:"title=Description,description=Human readable description of overlay" fingerprint:"-"`
 
+	// For overlays that target files inside a source archive, identifies the archive to modify.
+	// Must be a filename (not a path) matching a source archive in the component's sources directory.
+	// Only file-remove supports archive scoping; when set, it removes file(s) from inside the named
+	// archive instead of the loose sources tree.
+	Archive string `toml:"archive,omitempty" json:"archive,omitempty" jsonschema:"title=Archive,description=The source archive to modify (e.g. pkg-1.0.tar.gz)"`
```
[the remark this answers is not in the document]
Tonisal-byte: Same as below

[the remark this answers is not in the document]
Tonisal-byte: We aren't going to add this/ doesn't make sense

internal/projectconfig/overlay.go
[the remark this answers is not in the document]
Tonisal-byte: The above comment should really be here

internal/app/azldev/core/sources/archiveoverlays.go
```diff
@@ -0,0 +1,238 @@
...
+) error {
+	archivePath := filepath.Join(sourcesDirPath, group.archive)
+
+	// Create a temporary directory for extraction directly on the real filesystem.
+	// The [archive] package operates exclusively through OS primitives ([os.Root],
+	// os.*), so the work directory must be a genuine on-disk path regardless of the
+	// injected FS implementation. Using os.MkdirTemp here (instead of the injected
+	// FS) makes that requirement explicit and keeps the path valid even when fs is
+	// an in-memory or otherwise non-OS-backed FS (e.g., in tests or alternate runners).
+	workDir, err := os.MkdirTemp("", "archive-overlay-")
```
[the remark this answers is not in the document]
Tonisal-byte: Dry run does not belong at the overlay application layer, none of the other overlays honor dry-run either.

internal/app/azldev/core/sources/archiveoverlays.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	defer func() {
+		if removeErr := os.RemoveAll(workDir); removeErr != nil {
+			slog.Warn("Failed to clean up archive work directory", "error", removeErr)
+		}
+	}()
+
+	// Extract the archive; compression is inferred from the filename extension.
+	if err := archive.ExtractAuto(archivePath, workDir); err != nil {
```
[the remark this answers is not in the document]
Tonisal-byte: Same as above

internal/app/azldev/core/sources/archiveoverlays.go
```diff
@@ -0,0 +1,238 @@
...
+	// which operate solely on the destination tree, so the extract-root FS is passed as
+	// both the source and destination FS — there is no component-source FS to read from.
+	for _, overlay := range group.overlays {
+		if err := applyNonSpecOverlay(dryRunnable, extractFS, extractFS, overlay); err != nil {
+			return fmt.Errorf("applying %#q operation:\n%w", overlay.Type, err)
+		}
+	}
+
+	// Deterministically repack the archive in-place, reusing the original compression.
+	if err := archive.CreateDeterministicArchiveAuto(archivePath, workDir); err != nil {
```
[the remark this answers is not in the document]
Tonisal-byte: Same as above

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -647,7 +686,19 @@ func (p *sourcePreparerImpl) updateSourcesFile(component components.Component, o
...
+	// on the parsed representation — single parse, single write.
+	existingLines, err := fedorasource.ReadSourcesFile(existingContent)
+	if err != nil {
+		return fmt.Errorf("failed to parse 'sources' file %#q:\n%w", sourcesFilePath, err)
+	}
+
+	// Rehash archives that were modified by archive overlays in-place.
+	if err := p.rehashModifiedEntries(existingLines, outputDir, modifiedArchives); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Tonisal-byte: Same as above

docs/user/reference/config/overlays.md
Tonisal-byte: You're right, the comment was wrong. I make sure the comment is correct in the final version.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -647,7 +692,19 @@ func (p *sourcePreparerImpl) updateSourcesFile(component components.Component, o
...
```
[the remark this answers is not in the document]
Tonisal-byte: Same as above comment

internal/app/azldev/core/sources/archiveoverlays_internal_test.go
```diff
@@ -0,0 +1,171 @@
...
+			Type:     projectconfig.ComponentOverlayRemoveFile,
+			Filename: "*.conf",
+		}
+
+		err = applyNonSpecOverlay(ctx, ctx.FS(), extractFS, overlay)
+		require.NoError(t, err)
+
+		assert.FileExists(t, extractRoot+"/keep.txt")
+		assert.NoFileExists(t, extractRoot+"/remove.conf")
+	})
```
PawelWMS: I don't get this test. Does it test anything introduced in `archiveoverlays.go`? I only see a call to `applyNonSpecOverlay`, which existed for a while and is already tested elsewhere (I hope).

internal/app/azldev/core/sources/archiveoverlays_internal_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		root, err := resolveExtractRoot(workDir)
+		require.NoError(t, err)
+		assert.Equal(t, workDir, root)
+	})
+}
+
+// TestArchiveFileRemove verifies that archive-scoped file-remove overlays are
+// routed through the shared [applyNonSpecOverlay] machinery against the
+// extract-root FS (i.e., the same code path that [processArchive] uses).
```
PawelWMS: 1. Why do we test if we're re-using `applyNonSpecOverlay`? Seems like internal implementation detail and I'm not sure, it's a specific requirement we want to impose. Re-using existing code is definitely a good thing, though, I'm not questioning that.
2. If we do want to check for this specific case, however, the name `TestArchiveFileRemove` is too broad and suggests we're testing the removal of files inside archives, not re-use of `applyNonSpecOverlay`.
3. I don't think any of the tests in this function actually check we're calling `applyNonSpecOverlay` from the archive overlays logic.

docs/user/reference/config/overlays.md
```diff
@@ -274,6 +303,34 @@ description = "Remove CVE patches that are now upstream"
...
 
+### Removing a File from an Archive
+
+Prefix the `file` path with the archive name to delete files matching a glob pattern from inside a
+source archive. The archive is extracted, matching files are removed, and the archive is repacked.
+
+```toml
+[[components.mypackage.overlays]]
+type = "file-remove"
+file = "mypackage-1.0.tar.gz/vendor/**"
```
PawelWMS: I couldn't find tests for globs to make sure the following cases are treated correctly:
- `<archive>/some_folder` - removes the whole `some_folder` from inside the archive.
- `<archive>/some_folder/*` - removes the contents of `some_folder` inside the archive but keeps the folder itself.
- `<archive>/some_folder/**` - same as above.
- `<archive>/*/some_file.txt` - removes `some_file.txt` from **TOP LEVEL** folders inside the archive.
- `<archive>/**/some_file.txt` - removes `some_file.txt` from **ALL** folders inside the archive regardless of depth.
- Repetition of the cases above, but instead of `<archive>` we have a local checked-in directory.

internal/app/azldev/core/sources/archiveoverlays.go
```diff
@@ -0,0 +1,254 @@
...
+// full, permission error, etc.), it repacks to a temp file in the same directory
+// and atomically renames it over the original only on success. Repacking directly
+// over archivePath would truncate it first, leaving the workspace unrecoverable
+// without refetching if the repack then failed.
+func repackArchiveAtomic(archivePath, archiveName, workDir string) (err error) {
+	comp, err := archive.DetectCompression(archiveName)
+	if err != nil {
+		return fmt.Errorf("detecting compression for %#q:\n%w", archiveName, err)
+	}
+
```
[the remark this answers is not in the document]
Tonisal-byte: We aren't trying to preserve the original archive; won't fix

internal/app/azldev/core/sources/archiveoverlays_internal_test.go
dmcilvaney: +1

Ideally test behavior should in non-internal tests whenever possible (black box) as well.

internal/app/azldev/cmds/component/preparesources.go
```diff
@@ -194,3 +188,23 @@ func CheckOutputDir(env *azldev.Env, options *PrepareSourcesOptions) error {
 			"use --force to delete and recreate it",
 		options.OutputDir)
 }
+
+// appendPrepareSourcesOptions appends conditional preparer options that control
+// hashing and lookaside behavior. Extracted from
+// [PrepareComponentSources] to keep cyclomatic complexity within limits.
+func appendPrepareSourcesOptions(
```
dmcilvaney, line 194: I'd skip the editoralizing, a reader probably doesn't care about cyclomatic complexity when they are trying to read the source code.
```suggestion
// appendPrepareSourcesOptions appends conditional preparer options that control
// hashing and lookaside behavior.
func appendPrepareSourcesOptions(
```

internal/projectconfig/overlay.go
```diff
@@ -119,10 +121,48 @@ func (c *ComponentOverlay) ModifiesSpec() bool {
...
+
+	return ok
+}
+
 // ModifiesNonSpecFiles returns true if the overlay modifies non-spec files. This includes
 // hybrid overlays that modify both spec and source files (e.g., patch overlays), since
-// those also require non-spec modifications.
+// those also require non-spec modifications. Archive-scoped overlays (see [ModifiesArchive])
+// are excluded: they operate on files inside an archive, not loose files in the sources tree.
 func (c *ComponentOverlay) ModifiesNonSpecFiles() bool {
```
dmcilvaney: This func name might need a change to something like `ModifiesLooseFiles` or something. Real head scratcher when trying to trace the logic around.

internal/utils/archive/archive.go
```diff
@@ -174,6 +233,19 @@ func (r readerCloser) Close() error {
 	return nil
 }
 
+// CreateDeterministicArchiveAuto is a convenience wrapper that infers the
+// compression from archivePath's extension via [DetectCompression] and then
+// calls [CreateDeterministicArchive]. Most callers should prefer this over the
+// explicit-compression [CreateDeterministicArchive].
+func CreateDeterministicArchiveAuto(archivePath, sourceDir string) error {
```
dmcilvaney: `extractEntry()` skips some unexpected values in archives, should either make that a hard error, or very clearly document the expected behavior. There are a decent number of packages that internally use synlinks for example. (I think systemd tests for example have some really fun recursive ones)

Could gate the hard error to the overlay path if the existing call sites really don't care.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -666,6 +729,75 @@ func (p *sourcePreparerImpl) updateSourcesFile(component components.Component, o
 	return nil
 }
 
+// rehashModifiedEntries updates the Raw and Entry fields of parsed 'sources' lines
+// for archives that were modified by archive overlays. The hash is recomputed using
+// the same hash type as the original entry. It returns an error if any modified
+// archive has no matching 'sources' entry, since that would leave a stale digest.
+func (p *sourcePreparerImpl) rehashModifiedEntries(
```
dmcilvaney, line 736: Do we need to call this in `DiffSources` as well?

internal/app/azldev/cmds/component/preparesources.go
```diff
@@ -194,3 +188,23 @@ func CheckOutputDir(env *azldev.Env, options *PrepareSourcesOptions) error {
...
+	_ *azldev.Env,
+	opts []sources.PreparerOption,
+	options *PrepareSourcesOptions,
```
dmcilvaney, line 196: Do we even need all these args? Can drop the unused ones?

internal/utils/archive/archive.go
```diff
@@ -174,6 +233,19 @@ func (r readerCloser) Close() error {
...
+	comp, err := DetectCompression(archivePath)
+	if err != nil {
```
dmcilvaney: Also, is this going to be used anywhere? Currently seems unused.

internal/app/azldev/core/sources/archiveoverlays_internal_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	"github.com/microsoft/azure-linux-dev-tools/internal/global/testctx"
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileperms"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/rootfs"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGroupOverlaysByArchive(t *testing.T) {
```
dmcilvaney, line 22: Do we have an e2e test of this actually working with mutliple overlays against a single archive?

dmcilvaney: actaully, here is the detected test gaps as a list, decide if these are useful.

-  No end-to-end test for file-search-replace inside an archive — only file-remove is exercised through PrepareSources (search-replace is only asserted to route, never to rewrite+repack+rehash). High consensus.
-  No test for multi-overlay batching — the processArchive loop applying 2+ overlays to one archive in a single extract/repack cycle.
-  No hardlink / special-entry round-trip test (the Finding 1 data-loss path is completely uncovered).
-  No test driving the resolveExtractRoot fallback (single dir + sibling) through full processArchive.
-  No test for the --skip-sources archive-overlay skip-with-warning branch.
-  No test asserting empty-dir-survives / directory-removal-unsupported behavior.

internal/app/azldev/core/sources/sourceprep.go
Tonisal-byte: Probably not, it seems like DiffSources intentionally does not want to update its `sources` file

docs/user/reference/config/overlays.md
```diff
@@ -47,20 +47,51 @@ successfully makes a replacement to at least one matching file.
...
+> prefixing the `file` path with the archive name — see [Archive Overlays](#archive-overlays).
+
+### Archive Overlays
+
+A `file-remove` or `file-search-replace` overlay can modify files **inside** a source archive
+instead of loose files in the sources tree. This is detected from the `file` path: when its first
+segment is a source archive (e.g. `pkg-1.0.tar.gz`) followed by an inner path, the overlay is
+scoped to that archive and the remainder is matched against its contents. The archive is extracted,
+the matching files are modified with the same machinery as loose-file overlays, and the archive is
+repacked with its original compression format.
```
dmcilvaney, line 58: Might want to clarify this, I think we re-pack based on the extension, but extract based on magic bytes. (ie we might change the compression)

Tonisal-byte: Good catch, the archive are now repacked in their original format (even if it is a lie :'(

internal/utils/archive/archive.go
```diff
@@ -50,6 +57,11 @@ const (
...
 )
 
+// compressionMagicLen is the number of leading bytes [sniffCompression] needs
+// to recognize every supported compressed format (xz has the longest, 6-byte,
+// signature).
+const compressionMagicLen = 6
+
 // maxEntryBytes caps the decompressed size of any single regular-file entry
 // extracted by [Extract]. This prevents a decompression-bomb archive from
 // filling the destination filesystem. 10 GiB is well above any reasonable
```
dmcilvaney, line 67: Not related to this PR maybe, but max size of a single file is probably not enough to defend against decompression bombs, need to count full size of all files + inode use. Probably don't bother fixing here, but fyi.

internal/projectconfig/overlay.go
```diff
@@ -119,10 +121,50 @@ func (c *ComponentOverlay) ModifiesSpec() bool {
...
 
-// ModifiesNonSpecFiles returns true if the overlay modifies non-spec files. This includes
-// hybrid overlays that modify both spec and source files (e.g., patch overlays), since
-// those also require non-spec modifications.
-func (c *ComponentOverlay) ModifiesNonSpecFiles() bool {
+// ArchiveTarget inspects the overlay's file path and, when it targets files inside a source
+// archive, returns the archive filename and the glob to match within the extracted archive.
+//
+// An overlay is archive-scoped when the first path segment of [ComponentOverlay.Filename] is a
+// recognized archive name (see [archive.IsArchiveName]) and a non-empty inner path follows it,
```
dmcilvaney: wondering, would this logic be better:
If the path doesn't exist, and the root is a file, try to see if its an archive using the selection logic, repeat as needed.

Its possible that a directory might be called file.tar.gz

internal/projectconfig/overlay.go
dmcilvaney, line 330: Might want to add some sort of validation here for which types support archives (or if we go with the recursive loading flow error handling needs to happen at run time).

internal/projectconfig/overlay.go
Tonisal-byte: I don't think we want to add this yet, currently we don't want to support nested archives, but this could definitely be done in the future.

dmcilvaney (approved):

Tonisal-byte:
Already implemented this feature in a different PR

PR #225, opened 2026-06-03
ci: add python linters to cover future python code

dmcilvaney:

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,35 @@
+name: "Python lint and type-check"
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+  schedule:
+    # Run every night at 3:15am PST (11:15am UTC)
+    - cron: "15 11 * * *"
+
+# Cancel in-progress runs of this workflow if a new run is triggered.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  python:
+    name: "Lint and type-check"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Get go tools
+        uses: ./.github/getgotools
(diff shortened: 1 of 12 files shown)


.github/workflows/python.yml
```diff
@@ -0,0 +1,35 @@
+name: "Python lint and type-check"
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+  schedule:
+    # Run every night at 3:15am PST (11:15am UTC)
```
Tonisal-byte, line 9: Just curious, why do we need the linters to run on as schedule? Is it not enough to have them run on push?

Tonisal-byte (approved):
This will be very helpful (painful) for when I add azlpkg to the repo, approved!

.github/workflows/python.yml
dmcilvaney: We have it on the others, but probably don't truly need it. Mostly it catches if the linters themselves change.

reubeno (dismissed):
We have planned future changes to add Python tooling -- and we do already have some Python that's sneaked in -- totally understood. With that happening, what's the plan for how dev/AI tooling for a multi-language dev process?

This PR is titlted "add Python linters" but also bundles with it an implicit design of using `mage` as this tool--and I'd just like to make sure we're intentional around it.

If we end up building Python packages with pyproject.toml, etc. is it still the right design to use `mage` to automate that? If we are, does the Python stuff need to be separate verbs/targets, or bundled into the existing targets for `check`, `build`, `install`, etc.?

dmcilvaney:
> We have planned future changes to add Python tooling -- and we do already have some Python that's sneaked in -- totally understood. With that happening, what's the plan for how dev/AI tooling for a multi-language dev process?
> 
> This PR is titlted "add Python linters" but also bundles with it an implicit design of using `mage` as this tool--and I'd just like to make sure we're intentional around it.
> 
> If we end up building Python packages with pyproject.toml, etc. is it still the right design to use `mage` to automate that? If we are, does the Python stuff need to be separate verbs/targets, or bundled into the existing targets for `check`, `build`, `install`, etc.?

Mage is just a convenience layer since its already doing linting etc., all the tools are configured with native python configs that work fine on their own from the cmdline. If we move to pyproject.toml the configs can be migrated into those files if desired. I figured keeping it in mage makes life easier for the agents and I'm not super keen on maintaining two parallel build systems.

liunan-ms (approved):

PR #231, opened 2026-06-04
feat: in verbose mode have mock print all output

dmcilvaney:
This is VERY verbose, but it is very useful for tracking progress in pipelines etc.


internal/rpm/mock/mock.go
```diff
@@ -440,7 +440,11 @@ func (r *Runner) BuildRPM(ctx context.Context, srpmPath, outputDirPath string, o
 		return fmt.Errorf("failed to create external command for mock:\n%w", err)
 	}
 
-	extcmd = extcmd.SetLongRunning("Waiting for mock (building RPM)...")
+	// In verbose mode, mock streams its full build output to the console, so we skip the
+	// indeterminate progress spinner (which would otherwise fight with the live output).
+	if !r.verbose {
+		extcmd = extcmd.SetLongRunning("Waiting for mock (building RPM)...")
```
reubeno: Would it be cleaner to change what `.SetLongRunning()` does in verbose mode?

dmcilvaney: done

internal/rpm/mock/mock.go
Tonisal-byte: Nit (user experience): Wouldn't this result in all mock output being black/italic. Maybe if in verbose mode we have the output text be default?

Tonisal-byte (approved):

internal/rpm/mock/mock.go
dmcilvaney: updated

PR #232, opened 2026-06-04
ci: update CodeQL config

dmcilvaney:
In parallel to this we should ensure that CodeQL is configured in the repo settings.
I think we probably want to set it to fail on warnings (we are effectively green with one tiny exception, so may as well keep it that way).
- Ensure codeQL is configured to use the new yaml
- Enable Code quality (https://docs.github.com/en/code-security/concepts/about-code-quality)
- Set branch rules to require the check
- Set failure severity
- Fix outstanding issue in magefile completions.go


Tonisal-byte (approved):
LGTM

reubeno (approved):

PR #242, opened 2026-06-16
feat(component): add overlay metadata and 'component metadata' command

liunan-ms:
**Summary**
Overlays currently carry no record of why they exist or when they can be removed, so reviewers can't tell a temporary Fedora backport from a permanent AZL branding change. This PR adds an optional `[metadata]` block to overlays, a per-file `.overlay.toml` format for grouping related overlays, and a read-only `azldev component metadata` command for inspecting them.

**What's new**
1. Overlay metadata block
Overlays can now declare optional documentation metadata:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Field | Purpose
-- | --
category | Required. Classifies   intent (backport-fedora, upstream-fix, azl-branding-policy, …).
commits | Upstream commit   URLs this overlay backports/references.
pr | Upstream   pull-request URL.
bug | Issue-tracker   URLs.
upstreamability | yes / no / unknown (defaults to unknown).

</div>

</body>

</html>

2. Per-file .overlay.toml format (overlay-dir)

A component can set `overlay-dir = "overlays"` to load `*.overlay.toml` files, one per logical change. Each file has a single file-level `[metadata]` block (stamped onto every overlay in the file) plus an ordered list of `[[overlays]]`. Files load in filename order (0001-, 0002- convention), are appended after inline overlays, and resolve source paths relative to the overlay file. Per-overlay metadata inside these files is rejected.

3. Optional component-group metadata 

Component groups can now carry an optional `[metadata]` block that reuses the same schema as overlay metadata (category, commits, pr, bug, upstreamability). It's documentation-only: it doesn't affect how the group or its components are built and is excluded from fingerprints. The `azldev component metadata` command surfaces it alongside overlay metadata, tagged with `source = group`.

4. `azldev component metadata` command

A read-only command listing overlays and their metadata across components, with `--category`, `--only-annotated`, `--upstreamability` and `--overlays / --groups` filters. Table (default) or JSON output. Available as an MCP tool. Does not parse spec files or fetch sources, so it's fast and works with missing/stale locks.

**Validated against real overlays** 
```bash
# 1. Config loads and the overlays merge + validate (metadata rules enforced)
azldev component metadata -a --overlays

# 2. Inspect a single annotated component
azldev component metadata -p cpio --overlays

# 3. Exercise the filters
azldev component metadata -a --overlays --only-annotated
azldev component metadata -a --overlays --category backport-dist-git
azldev component metadata -a --overlays --upstreamability yes

# 4. metadata for component groups
azldev comp metadata -p gnutls --groups     # groups only

# 5. metadata declares category = "azl-test-disablement" from both sources 
azldev comp metadata -a --category azl-test-disablement
```


internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,157 @@
...
+	if !permissiveConfigParsing {
+		decoder.DisallowUnknownFields()
+	}
+
+	var ofile OverlayFile
+	if err := decoder.Decode(&ofile); err != nil {
+		return nil, fmt.Errorf("failed to parse overlay file %q:\n%w", overlayPath, err)
+	}
+
+	if err := ofile.Metadata.Validate(); err != nil {
```
dmcilvaney, line 188: I think this won't play well with https://github.com/microsoft/azure-linux-dev-tools/pull/216.

For some future work we need to be able to retroactively load old configs using the permissive config loading mode. Need to make sure `permissiveConfigParsing` is passed in and handled.

internal/projectconfig/overlay.go
```diff
@@ -333,6 +340,12 @@ func (c *ComponentOverlay) Validate() error {
 		return fmt.Errorf("unknown overlay type %#q: %#q", c.Type, desc)
 	}
 
+	if c.Metadata != nil {
+		if err := c.Metadata.Validate(); err != nil {
```
dmcilvaney, line 344: Need to handle permissive mode.

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,157 @@
...
+// resolves overlay `source` paths relative to the overlay file.
+func loadOverlayDir(
+	fs opctx.FS, absDir string, permissiveConfigParsing bool,
+) ([]ComponentOverlay, error) {
+	pattern := path.Join(absDir, overlayFileSuffix)
+
+	matches, err := fileutils.Glob(
+		fs, pattern,
+		doublestar.WithFailOnIOErrors(),
+		doublestar.WithFilesOnly(),
```
dmcilvaney, line 137: What happens if the glob matches nothing? Feels like that should be an error rather than an empty list. I think https://github.com/bmatcuk/doublestar/blob/a9ad9e0ef4d6b7e4443090e9a7201d847a881711/globoptions.go#L56 should do it.

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,157 @@
...
+
+// applyOverlayDirs scans each component's [ComponentConfig.OverlayDir] (when set),
+// parses every `*.overlay.toml` file in deterministic (filename) order, stamps the
+// per-file metadata onto each overlay, and appends the resulting overlays to the
+// component's [ComponentConfig.Overlays] slice. Inline overlays declared directly on
+// the component come first; file-sourced overlays are appended after them.
+//
+// Called from [loadProjectConfigFile] after TOML decode but before [ConfigFile.Validate],
+// so all per-overlay validation rules apply uniformly regardless of declaration site.
+func applyOverlayDirs(fs opctx.FS, cfg *ConfigFile, permissiveConfigParsing bool) error {
```
dmcilvaney, line 74: We merge in default configs (`ProjectConfig.DefaultComponentConfig`, `ComponentGroupConfig.DefaultComponentConfig`, `DistroDefinition.DefaultComponentConfig`, ConfigFile.DefaultComponentConfig`) after this occurs.

Practically, I think this may be ok, overlay dirs could be scoped to just specific components (although explicit in-line overlays DO apply from groups).

More generally though, it feels like this might need a rework in the future, we should probably have a single common path for loading that ensures validation happens at all the right points.

A stopgap fix could be to just fail validation on any default config that tries to set this.

internal/projectconfig/overlay_metadata.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// PR is a link to the upstream pull request that carries (or proposes) the fix.
+	PR string `toml:"pr,omitempty" json:"pr,omitempty" validate:"omitempty,http_url" jsonschema:"title=Upstream PR,description=URL of the upstream pull request"`
+
+	// BugLinks holds URLs of issue-tracker entries related to this overlay.
+	BugLinks []string `toml:"bug,omitempty" json:"bug,omitempty" validate:"omitempty,dive,http_url" jsonschema:"title=Issue-tracker links,description=URLs of related issue-tracker entries"`
+
+	// Upstreamability records whether this overlay's change can be upstreamed: 'yes', 'no',
+	// or 'unknown'. Omitting the field defaults to 'unknown' (not yet assessed).
+	Upstreamability OverlayUpstreamability `toml:"upstreamability,omitempty" json:"upstreamability,omitempty" jsonschema:"enum=yes,enum=no,enum=unknown,default=unknown,title=Upstreamability,description=Whether this overlay's change can be upstreamed: 'yes', 'no', or 'unknown' (the default)"`
```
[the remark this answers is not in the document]
dmcilvaney: ^ seconded.

internal/projectconfig/overlay_metadata.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+// OverlayMetadata describes the intent and provenance of an overlay. It is documentation
+// only — it does not affect how the overlay is applied and is excluded from component
+// fingerprints. When present, it must declare a [OverlayMetadata.Category]; other fields
+// are optional but constrained by category-specific rules (see [OverlayMetadata.Validate]).
+type OverlayMetadata struct {
+	// Category classifies the overlay's intent. Required.
+	Category OverlayCategory `toml:"category" json:"category" jsonschema:"required,enum=backport-fedora,enum=upstream-fix,enum=azl-dependency-pruning,enum=azl-feature-disablement,enum=azl-branding-policy,enum=azl-build,enum=azl-test-disablement,enum=azl-security-compliance,enum=azl-release-management,enum=azl-missing-dependency-workaround,enum=azl-platform-adaptation,title=Category,description=Classification of the overlay's intent"`
```
dmcilvaney: Intentionally omitted the `validate` tag? Can probably simplify a bunch of the validation logic later on.

internal/projectconfig/overlay_file.go
liunan-ms: `permissiveConfigParsing` is already threaded through `loadProjectConfigFile` -> `applyOverlayDirs` -> `loadOverlayDir` -> `loadOverlayFile`, where it controls `decoder.DisallowUnknownFields()` on the per-file TOML decoder, mirroring the project-config decoder. So a retroactive load against an older .overlay.toml with unknown fields should be tolerated in permissive mode just like the parent config.

internal/projectconfig/overlay_file.go
liunan-ms: Fixed. `loadOverlayDir` now passes `doublestar.WithFailOnPatternNotExist()` and also checks `len(matches) == 0`, returning a new `ErrOverlayDirNoFiles` sentinel. Also updated tests to cover both the empty-but-exists and missing-dir cases.

internal/projectconfig/overlay_file.go
liunan-ms: I agree this needs a real rework. As a stopgap, `applyOverlayDirs` now calls `rejectOverlayDirInDefaults`, which returns `ErrOverlayDirInDefaultConfig` if `overlay-dir` is set on `DefaultComponentConfig` at the project, component-group, or distro-version layer, so the misconfiguration is loud rather than silently dropped.

internal/projectconfig/overlay_metadata.go
liunan-ms: Due to metadata redesign, the category set was consolidated to a single `backport-dist-git`, dropping `FixedIn` and `RemovableAfter` along with their category-specific validation. The remaining string fields (PR, Commits, BugLinks) all carry validate:"...,http_url" (or dive,http_url) tags and are enforced via the shared metadataValidator instance.

internal/app/azldev/cmds/component/overlays.go
```diff
@@ -0,0 +1,234 @@
...
+		return false
+	}
+
+	if options.Upstreamability != "" {
+		// The filter value is validated upfront in ListOverlays, so the error is unreachable here.
+		want, _ := normalizeUpstreamabilityFilter(options.Upstreamability)
+		if info.Upstreamability != want {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
liunan-ms: fixed in [189edbd](https://github.com/microsoft/azure-linux-dev-tools/pull/242/commits/189edbde0f9ef1c14814e09f6e798298500336c5)

internal/projectconfig/overlay_metadata.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf(
+			"unknown upstreamability %#q; want %#q, %#q, or %#q",
+			string(m.Upstreamability), string(OverlayUpstreamabilityYes),
+			string(OverlayUpstreamabilityNo), string(OverlayUpstreamabilityUnknown),
+		)
+	}
+
+	// Field-level constraints (for example 'pr' and 'bug' must be http(s) URLs) are
+	// expressed as 'validate' struct tags and enforced here.
+	if err := validator.New().Struct(m); err != nil {
```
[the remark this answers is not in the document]
liunan-ms: fixed in [189edbd](https://github.com/microsoft/azure-linux-dev-tools/pull/242/commits/189edbde0f9ef1c14814e09f6e798298500336c5)

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,157 @@
...
+	overlayDir := filepath.Dir(overlayPath)
+
+	for idx := range ofile.Overlays {
+		overlay := &ofile.Overlays[idx]
+
+		if overlay.Metadata != nil {
+			return nil, fmt.Errorf("%w (overlay %d in %q)", ErrOverlayFilePerOverlayMetadata, idx+1, overlayPath)
+		}
+
+		overlay.Metadata = deep.MustCopy(&ofile.Metadata)
```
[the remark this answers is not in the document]
liunan-ms: fixed in [189edbd](https://github.com/microsoft/azure-linux-dev-tools/pull/242/commits/189edbde0f9ef1c14814e09f6e798298500336c5)

internal/projectconfig/overlay_metadata.go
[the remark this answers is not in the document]
liunan-ms: fixed in [189edbd](https://github.com/microsoft/azure-linux-dev-tools/pull/242/commits/189edbde0f9ef1c14814e09f6e798298500336c5)

internal/projectconfig/overlay_file.go
dmcilvaney: Looking more, I think there is still an issue here though, an old config could set something that a new version of Metadata.Validate() doesn't like (ie we deprecate one of the flags etc.).

To align with #216 here we would need to take the error from `Metadata.Validate()` and print it as a warning, then continue like `loader.go:62`)

liunan-ms: Per-file overlay metadata validation now warns and continues under
--permissive-config instead of failing the load, aligning with the
behavior introduced in #216.

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,218 @@
...
+	overlayDir := filepath.Dir(overlayPath)
+
+	for idx := range ofile.Overlays {
+		overlay := &ofile.Overlays[idx]
+
+		if overlay.Metadata != nil {
+			return nil, fmt.Errorf("%w (overlay %d in %q)", ErrOverlayFilePerOverlayMetadata, idx+1, overlayPath)
+		}
+
+		overlay.Metadata = ofile.Metadata.clone()
```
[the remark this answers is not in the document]
liunan-ms: The suggestion breaks the restriction of one [metadata] per file. Fixing ConfigFile.Validate() / ComponentOverlay.Validate() will be added in future work.

dmcilvaney (approved):

PR #249, opened 2026-06-24
feat(overlays): add overlay metadata and per-file overlay format (1/3)

liunan-ms:
This PR is the first of three-split out from the previous overlay-metadata PR (#242). It adds two related ways to document overlays without affecting how they're applied. The follow-ups will add component-group metadata (PR 2) and the `azldev component metadata` command that consumes both (PR 3).

**What's new**

- Inline [metadata] block on overlays (commit 1)
 Optional table on any overlay with:

  - category (required) — one of backport-dist-git, azl-pruning, azl-compatibility, azl-branding-policy, azl-dep-missing-workaround, azl-disable-flaky-tests, azl-disable-unsupported-tests, azl-security-compliance, azl-release-management, azl-platform-adaptation
  - commits — upstream commit URLs (required when category = "backport-dist-git")
  - bugs — bug-tracker URLs
  - upstreamable — true/false; omit when not yet assessed

- Per-file overlay format via overlay-files (commit 2)
For logical changes that span multiple overlays (CVE backports, feature disablements, etc.), set overlay-files on a component to one or more globs. Each matched file declares a single file-level [metadata] block (stamped onto every overlay in the file) plus an ordered list of [[overlays]].
Files are loaded in filename order within each glob, globs concatenated in declaration order, duplicates de-duped. Per-overlay [metadata] is rejected inside an overlay file (file-level metadata is the single source of truth). Relative source paths resolve against the overlay file's directory. An empty glob match is a configuration error.

**Notes**
- Metadata and overlay-files are excluded from component fingerprints — editing them never invalidates build caches.
- Regenerated JSON schema and scenario snapshots.
- No CLI surface yet; `azldev component metadata` consumes this in a follow-up PR.


internal/projectconfig/overlay_metadata.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return &cloned
+}
+
+// metadataValidator is a shared, concurrency-safe validator instance reused across all
+// [OverlayMetadata.Validate] calls to avoid per-call allocation.
+//
+//nolint:gochecknoglobals // validator instances are safe for concurrent use once configured.
+var metadataValidator = validator.New()
+
```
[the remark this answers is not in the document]
liunan-ms: I re-checked this against the repo-selected dependency. `go.mod` pins `github.com/go-playground/validator/v10 v10.30.3`, and that version includes `http_url` as a built-in validation tag (`"http_url": isHttpURL` in `baked_in.go`). The overlay metadata tests also exercise this path; if the tag were unknown, `validator.Struct` would fail/panic instead of returning the expected URL validation errors. No code change needed here.

internal/projectconfig/overlay.go
```diff
@@ -43,6 +43,13 @@ type ComponentOverlay struct {
 	// Excluded from fingerprint because it contains an absolute path that varies by checkout
 	// location. Overlay content is hashed separately by [fingerprint.ComputeIdentity].
 	Source string `toml:"source,omitempty" json:"source,omitempty" jsonschema:"title=Source,description=For overlays that require a source file as input, indicates a path to that file; relative paths are relative to the config file that defines the overlay" fingerprint:"-"`
+
+	// Metadata describes the intent and provenance of the overlay (category, related
+	// commits/PR, bug links, etc.). Optional. Populated either inline in the component
+	// config file or by the [ComponentConfig.OverlayFiles] loader, which stamps the
+	// per-file metadata onto every overlay declared in that file. Excluded from the
+	// fingerprint because it is documentation only.
```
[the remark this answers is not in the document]
liunan-ms: fixed

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,239 @@
...
+
+	for idx := range ofile.Overlays {
+		overlay := &ofile.Overlays[idx]
+
+		if overlay.Metadata != nil {
+			return nil, fmt.Errorf("%w (overlay %d in %q)", ErrOverlayFilePerOverlayMetadata, idx+1, overlayPath)
+		}
+
+		overlay.Metadata = ofile.Metadata.clone()
+		overlay.Source = makeAbsolute(overlayDir, overlay.Source)
```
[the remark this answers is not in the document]
liunan-ms: fixed

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,245 @@
...
+		switch {
+		case errors.Is(err, doublestar.ErrPatternNotExist):
+			return nil, fmt.Errorf("%w: %q", ErrOverlayFilesNoMatches, pattern)
+		case err != nil:
+			return nil, fmt.Errorf("failed to scan for overlay files matching %q:\n%w", pattern, err)
+		case len(matches) == 0:
+			return nil, fmt.Errorf("%w: %q", ErrOverlayFilesNoMatches, pattern)
+		}
+
+		sort.Strings(matches)
```
dmcilvaney: Not sure it matters, all files should have the same directory structure, but as written in the docs it should sort by basename.

internal/projectconfig/component.go
```diff
@@ -273,6 +273,19 @@ type ComponentConfig struct {
...
+	// Each matched file is parsed as an [OverlayFile]: one logical change consisting of a
+	// file-level `[metadata]` block plus an ordered list of `[[overlays]]`. The per-file
+	// metadata is stamped onto every overlay in the file. Matches are concatenated in the
+	// order patterns are declared; within a single pattern, matches are applied in
+	// filename (lexicographic) order, using the full path as a tie-breaker when
+	// filenames match. Duplicate matches are de-duplicated, preserving first
+	// occurrence. The resulting overlays are appended to [ComponentConfig.Overlays]
+	// after any inline overlays. Excluded from the fingerprint because the value affects
+	// only where overlays are sourced from, not their content.
+	OverlayFiles []string `toml:"overlay-files,omitempty" json:"overlayFiles,omitempty" table:"-" jsonschema:"title=Overlay files,description=Glob patterns (relative to the component config file) matched against the filesystem to locate per-file overlay documents at load time" fingerprint:"-"`
```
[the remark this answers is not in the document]
liunan-ms: fixed

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,253 @@
...
+func rejectOverlayFilesInDefaults(cfg *ConfigFile) error {
+	if cfg.DefaultComponentConfig != nil && len(cfg.DefaultComponentConfig.OverlayFiles) > 0 {
+		return fmt.Errorf("%w (project-level default-component-config)", ErrOverlayFilesInDefaultConfig)
+	}
+
+	for name, group := range cfg.ComponentGroups {
+		if len(group.DefaultComponentConfig.OverlayFiles) > 0 {
+			return fmt.Errorf("%w (component-group %#q)", ErrOverlayFilesInDefaultConfig, name)
+		}
+	}
```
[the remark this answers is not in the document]
liunan-ms: added table-driven coverage for rejecting overlay-files in project-level, component-group, distro configs.

internal/projectconfig/overlay_file.go
liunan-ms: Thanks for catching this! Updated to align with the docs, sort files in filename lexicographic order.

internal/projectconfig/loader.go
```diff
@@ -433,6 +433,13 @@ func loadProjectConfigFile(
 	cfg.sourcePath = absFilePath
 	cfg.dir = filepath.Dir(absFilePath)
 
+	// Resolve per-component overlay file globs, stamping each
+	// file's [metadata] onto its overlays and appending them to the component before
+	// validation runs.
+	if err := applyOverlayFiles(fs, cfg, permissiveConfigParsing); err != nil {
```
reubeno, line 439: This seems out-of-place as a special case in the toplevel loader function. Is there a way we could have done this closer to where the component / overlays are being resolved/loaded?

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,253 @@
...
+// patterns (when set), parses every matched file as an [OverlayFile] in deterministic
+// order, stamps the per-file metadata onto each overlay, and appends the resulting
+// overlays to the component's [ComponentConfig.Overlays] slice. Inline overlays
+// declared directly on the component come first; file-sourced overlays are appended
+// after them.
+//
+// Called from [loadProjectConfigFile] after TOML decode but before [ConfigFile.Validate],
+// so all per-overlay validation rules apply uniformly regardless of declaration site.
+func applyOverlayFiles(fs opctx.FS, cfg *ConfigFile, permissiveConfigParsing bool) error {
+	if err := rejectOverlayFilesInDefaults(cfg); err != nil {
```
reubeno, line 74: Why do we need to do this?

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,253 @@
...
+	for _, pattern := range patterns {
+		absPattern := pattern
+		if !filepath.IsAbs(absPattern) {
+			absPattern = path.Join(referenceDir, absPattern)
+		}
+
+		matches, err := fileutils.Glob(
+			fs, absPattern,
+			doublestar.WithFailOnIOErrors(),
+			doublestar.WithFailOnPatternNotExist(),
```
reubeno, line 149: Should we fail on the pattern not existing? That prevents the ability for a given TOML project to have a defaulted glob pattern that conditionally applies to every component.

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,253 @@
...
+		slog.Warn(
+			"Overlay file metadata validation failed; dropping metadata and continuing due to '--permissive-config'",
+			"overlayFile", overlayPath,
+			"error", err,
+		)
+
+		stampedMetadata = nil
+	}
+
+	if len(ofile.Overlays) == 0 {
```
reubeno, line 235: This is an odd occurrence -- but is there a reason to explicitly look for and fail on it?

internal/projectconfig/overlay_file.go
```diff
@@ -0,0 +1,253 @@
...
+	var ofile OverlayFile
+	if err := decoder.Decode(&ofile); err != nil {
+		return nil, fmt.Errorf("failed to parse overlay file %q:\n%w", overlayPath, err)
+	}
+
+	// In permissive mode, drop invalid file-level metadata so overlays still load
+	// cleanly; otherwise stamping the bad metadata would cause overlay-level
+	// validation to re-fail later with a misleading location.
+	stampedMetadata := &ofile.Metadata
+	if err := ofile.Metadata.Validate(); err != nil {
```
reubeno, line 221: Is there additional overlay validation applied to the inline overlays that we should applying here too? (I forget.)

reubeno (commented):
Generally looks pretty good to me -- just a few questions/comments.

internal/projectconfig/loader.go
liunan-ms: I considered moving this closer to component resolution, but it would require passing `permissiveConfigParsing` into the merge/resolve path, which currently only works on already-decoded config structs. Daniel will clean up the loading stuff, and we can move this closer to component resolution then.

internal/projectconfig/overlay_file.go
liunan-ms: Currently, we don't support `overlay-file` set in DefaultComponentConfig because every component has its own `overlays/*.overlay.toml`, all inherited patterns should be expanded relative to component’s config file. To support defaulted patterns, overlay file expansion has to be moved after a component’s effective config is resolved. I'm still thinking about some questions like how to avoid applying the same overlay twice if a component is resolved multiple times. I'd keep this PR rejecting overlay-files in defaults until I have a better design.
Here is some discussion in pervious PR: https://github.com/microsoft/azure-linux-dev-tools/pull/242#discussion_r3423590575

internal/projectconfig/overlay_file.go
liunan-ms: Same reason as above.

internal/projectconfig/overlay_file.go
liunan-ms: An overlay file is supposed to represent one logical change, and a logical change with zero overlays is probably a typo or stale file, so failing avoids silent no-op overlay files.

internal/projectconfig/overlay_file.go
liunan-ms: Yes, today file-loaded overlays get the same `ComponentOverlay.Validate()` pass indirectly: `applyOverlayFiles` appends them to `component.Overlays`, and then `ConfigFile.Validate()` validates the combined inline + file-loaded overlay list. The explicit validation here is only for the file-level `[metadata]` block before it is stamped onto each overlay.

internal/projectconfig/loader.go
reubeno: I think it's worth making that change -- I see this as adding debt; I'm okay if you defer to a later PR, but my request is that you file a bug on it and assign to yourself ( or Daniel if he's willing to take it) to make sure this doesn't get forgotten.

reubeno (approved):

PR #250, opened 2026-06-24
build(deps): upgrade go-toml from v2.4.0 to v2.4.2

liunan-ms:
Upgrades `github.com/pelletier/go-toml/v2` from `v2.4.0` to `v2.4.2`.

## Changes in go-toml v2.4.2:
- **v2.4.1**: Fix multiline string, array indentation, and omitempty behavior
- **v2.4.2**: Fix decode of embedded structs tagged with options only (e.g. `,inline`)

## Testing

- `mage unit` ✅
- `mage check all` ✅

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/muesli/termenv v0.16.0
 	github.com/nxadm/tail v1.4.11
 	github.com/opencontainers/selinux v1.15.1
-	github.com/pelletier/go-toml/v2 v2.4.0
+	github.com/pelletier/go-toml/v2 v2.4.2
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/samber/lo v1.53.0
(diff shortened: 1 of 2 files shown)


reubeno (commented):
It looks like https://github.com/microsoft/azure-linux-dev-tools/pull/248 (with copilot + Daniel's help) took part of the upgrade, with a snapshot fix to address scenario test failures.

What do you think about taking that in first, and then rebasing this atop it afterwards?

liunan-ms:
> It looks like #248 (with copilot + Daniel's help) took part of the upgrade, with a snapshot fix to address scenario test failures.
> 
> What do you think about taking that in first, and then rebasing this atop it afterwards?

Yea, I'll wait for #248 merged in and continue upgrading to v2.4.2 as this version includes a couple of fixes.

trungams (approved):

PR #256, opened 2026-06-26
fix(overlays): expand overlay files after config resolution

### LINKED ISSUE #253: Move overlay-files expansion closer to component resolution
**Description**

`overlay-files` are currently expanded during individual config file loading, in` loadProjectConfigFile()` via `applyOverlayFiles()`(introduced by #249). 

Overlay-files expansion should be moved out of the top-level config file loader and make it closer to component resolution without threading `permissiveConfigParsing` through merge/validation functions.

liunan-ms:
**Summary**

This updates `overlay-files` expansion so per-file overlay documents are loaded after component config resolution instead of during raw config file loading.

That means `overlay-files` can now be declared in `default-component-config` at the project, distro, or component-group level and inherited by every applicable component. Relative patterns are interpreted for each concrete component, so a shared default such as:

```toml
[default-component-config]
overlay-files = ["overlays/*.overlay.toml"]
```
can discover component-local overlay files for each component.

**Behavior Changes**

- Allows overlay-files in default component config.
- Applies inherited overlay-files to every component that receives that default.
- Resolves inherited relative overlay-file globs relative to the concrete component config file, or the matched spec directory for spec-discovered components.
- Treats overlay-file glob patterns with no matches as no-op instead of an error.
- Clears OverlayFiles after expansion so overlays are not appended more than once.
- Preserves component-level override behavior: a non-empty component overlay-files list replaces the inherited list.

Fixed #253

### DIFF
--- a/docs/user/reference/config/overlays.md
+++ b/docs/user/reference/config/overlays.md
@@ -72,7 +72,7 @@ In addition to per-overlay fields, the following fields are set directly on the 
 
 | Field | TOML Key | Description |
 |-------|----------|-------------|
-| Overlay files | `overlay-files` | List of glob patterns (relative to the component's config file) matched against the filesystem at load time to locate per-file overlay documents. Patterns support `**` (globstar). Matches are concatenated in declaration order; within a single pattern, matches are applied in filename (lexicographic) order, with full path as a tie-breaker for duplicate filenames. Each pattern must match at least one file. Duplicate matches across patterns are de-duplicated. See [Per-file overlay format](#per-file-overlay-format). |
+| Overlay files | `overlay-files` | List of path or glob patterns matched against the filesystem after component config resolution to locate per-file overlay documents. Relative patterns are resolved from the concrete component's config file, or from the matched spec file's directory for spec-discovered components. Patterns support `**` (globstar). Matches are concatenated in declaration order; within a single pattern, matches are applied in filename (lexicographic) order, with full path as a tie-breaker for duplicate filenames. Glob patterns that match no files are ignored; literal paths must match a file. Duplicate matches across patterns are de-duplicated. See [Per-file overlay format](#per-file-overlay-format). |
 
 ## Overlay Metadata
 
@@ -163,6 +163,8 @@ When a single logical change (a CVE backport, a feature disablement, a Fedora ch
 
 Set `overlay-files` on the component to one or more globs (relative to the component config) and drop one overlay file per logical change into a directory of your choosing. The conventional layout uses a sibling `overlays/` directory and a `*.overlay.toml` filename suffix, but neither is required — `overlay-files` is just a glob, so any layout you can describe with `**`/`*` patterns works.
 
+`overlay-files` can also be inherited from `default-component-config` at the project, distro, or component-group level. Inherited relative patterns are still resolved for each concrete component: from its component config file when it has one, or from the matched spec file's directory when it is discovered by a component group's `specs` pattern. This makes defaults useful for component-local discovery patterns such as `overlay-files = ["overlays/*.overlay.toml"]`. If a component sets `overlay-files`, that value replaces the inherited list; use `overlay-files = []` to disable inherited overlay files for a component, or include both patterns explicitly when you want to keep default discovery and add component-specific locations.
+
 ```
 base/comps/mypackage/
 ├── mypackage.comp.toml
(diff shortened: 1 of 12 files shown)


internal/app/azldev/core/components/resolver.go
```diff
@@ -459,7 +461,9 @@ func (r *Resolver) getComponentFromNameAndSpecPath(name, specPath string) (compo
 	var updatedComponentConfig *projectconfig.ComponentConfig
 
 	// Apply inherited defaults to the component.
-	updatedComponentConfig, err = applyInheritedDefaultsToComponent(r.env, foundComponentConfig)
+	overlayFilesReferenceDir := overlayFilesReferenceDir(foundComponentConfig, specPath)
+
+	updatedComponentConfig, err = applyInheritedDefaultsToComponent(r.env, foundComponentConfig, overlayFilesReferenceDir)
```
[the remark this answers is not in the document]
liunan-ms: fixed

internal/projectconfig/overlay_file.go
```diff
@@ -146,17 +113,14 @@ func loadOverlayFiles(
...
 		)
 
 		switch {
-		case errors.Is(err, doublestar.ErrPatternNotExist):
-			return nil, fmt.Errorf("%w: %q", ErrOverlayFilesNoMatches, pattern)
 		case err != nil:
 			return nil, fmt.Errorf("failed to scan for overlay files matching %q:\n%w", pattern, err)
 		case len(matches) == 0:
-			return nil, fmt.Errorf("%w: %q", ErrOverlayFilesNoMatches, pattern)
+			continue
```
Tonisal-byte, line 131: Should a no-match here really be silent? We dropped ErrOverlayFilesNoMatches, so a typo'd literal path (e.g. overlasy/foo.toml) now applies nothing with no error. Could we keep tolerating real globs but still error on literal paths that match nothing, like containsPattern does for includes in loader.go?

Tonisal-byte (commented):
Good work! Just one small comment

docs/user/reference/config/overlays.md
```diff
@@ -163,6 +163,8 @@ When a single logical change (a CVE backport, a feature disablement, a Fedora ch
 
 Set `overlay-files` on the component to one or more globs (relative to the component config) and drop one overlay file per logical change into a directory of your choosing. The conventional layout uses a sibling `overlays/` directory and a `*.overlay.toml` filename suffix, but neither is required — `overlay-files` is just a glob, so any layout you can describe with `**`/`*` patterns works.
 
+`overlay-files` can also be inherited from `default-component-config` at the project, distro, or component-group level. Inherited relative patterns are still resolved for each concrete component: from its component config file when it has one, or from the matched spec file's directory when it is discovered by a component group's `specs` pattern. This makes defaults useful for component-local discovery patterns such as `overlay-files = ["overlays/*.overlay.toml"]`. If a component sets a non-empty `overlay-files` list, that list replaces the inherited list; include both patterns explicitly when you want to keep default discovery and add component-specific locations.
```
dmcilvaney: Not your bug but probably should be fixed either here or with a fast follow PR.

Tracing the flow, I think component groups may not work. `GroupsByComponent` is populated from a group's `Componets` list. `GetComponentGroupByName ` finds glob'd specs, but aren't written back, and `applyInheritedDefaultsToComponent` uses the `GroupsByComponent[name]` map, so a glob'd spec won't get the new overlays.

internal/projectconfig/overlay_file.go
dmcilvaney: Agree here, we should at least emit a warning, but ideally there is some way to distinguish between globs that we expect to fail, vs. "real" entries.

internal/projectconfig/component.go
```diff
@@ -317,11 +318,17 @@ var AllowedSourceFilesHashTypes = map[fileutils.HashType]bool{
...
 // Mutates the component config, updating it with overrides present in other.
 func (c *ComponentConfig) MergeUpdatesFrom(other *ComponentConfig) error {
+	otherOverlayFiles := slices.Clone(other.OverlayFiles)
+
 	err := mergo.Merge(c, other, mergo.WithOverride, mergo.WithAppendSlice)
 	if err != nil {
 		return fmt.Errorf("failed to merge project info:\n%w", err)
 	}
 
+	if len(otherOverlayFiles) > 0 {
```
dmcilvaney: Should we have a way to make `overlay-files = []` in a component override the inherited values? ie we don't want them for some specific package.

internal/projectconfig/overlay_file.go
liunan-ms: Thanks, fixed. overlay-files now distinguishes literal paths from glob patterns after expansion: missing literal paths return an error, while glob patterns with no matches remain a no-op.

docs/user/reference/config/overlays.md
liunan-ms: Good catch. I fixed this in the resolver by carrying dynamic component-group membership for spec-discovered components into inherited default resolution. I also found and fixed the related direct SpecPaths filter case, where a component resolved by spec path could still miss group defaults.

internal/projectconfig/component.go
liunan-ms: Fixed. overlay-files = [] now explicitly clears inherited overlay-file patterns, while an omitted value still inherits lower-priority defaults and a non-empty value replaces them.

dmcilvaney (approved):

PR #257, opened 2026-06-28
fix: Mitigate render-drift for archive-overlays

Tonisal-byte:
We need a way to track the resulting hash from the sum of archive overlays applied to a source. This is a blocking issue for adding archive-overlays to upstream; otherwise the render checks fail from having a stale `sources` file pre-archive-overlay. 

Example of new origin-stub:
```toml
[[components.apache-commons-compress.source-files]]
filename = "commons-compress-1.27.1-src.tar.gz"
hash = "c7a2cef26959e687ad19b96b5ba8393d7514095e13bf0f29bd41e6b3c3cb2260d8ff23283ff3d5fd137b2522b843e7f0f50ab46bcf0f66df5383674f35f223ab"
hash-type = "SHA512"
origin = { type = "overlay"}
replace-upstream = true
replace-reason = "Upstream source tarball contains test fixtures flagged as malware by the AZL RPM signing pipeline. These files are not needed at runtime and are removed to allow SRPM publication."

```


Tonisal-byte:
Isn't necessary with the latest archive overlay functionality

PR #261, opened 2026-06-30
feat(source-files): add custom origin type for mock-based source generation

Tonisal-byte:
Introduces a new `origin.type = "custom"` for component source-files entries. Instead of downloading a pre-built artifact, azldev runs a user-supplied shell script inside a fresh mock chroot and packages the output directory as a deterministic archive.

Example:
```toml
[[components.yara.source-files]]
filename         = "yara-4.5.4.tar.gz"
origin.type           = "custom"
origin.script           = "modify_source.sh"
origin.mock-packages    = ["curl", "tar", "gzip", "coreutils"]
hash-type        = "SHA512"
hash             = "somehash"
```

Script contract: the script must write all desired archive contents to `/azldev-gen/output/`. azldev mounts the script read-only at /azldev-gen/script/<name> and the output directory read-write at /azldev-gen/output/, then packages the output into the declared filename using CreateDeterministicArchive.

Example:
```bash 
echo "[N/N] Repacking deterministically as ${MODIFIED_NAME}"
# azldev packages /azldev-gen/output/ into a deterministic archive.
# Place the stripped source tree there so the archive contains the same
# yara-4.5.4/ top-level directory the upstream tarball had.
mv "${EXTRACTED_DIRNAME}" "/azldev-gen/output/${EXTRACTED_DIRNAME}"
```


internal/providers/sourceproviders/sourcemanager.go
```diff
@@ -260,6 +280,48 @@ func (m *sourceManager) createComponentProviders(distro ResolvedDistro) {
...
+// has a 'mock-config' path configured, enabling 'custom' origin source files for
+// all commands without per-command wiring.
+//
+// Failures are logged and the provider is skipped, matching the tolerant
+// registration pattern used by [createComponentProviders].
+func (m *sourceManager) createFileProviders(env *azldev.Env) {
+	_, distroVerDef, err := env.Distro()
+	if err != nil {
+		slog.Debug("Cannot resolve project distro; 'custom' origin source generation will be unavailable",
+			"error", err)
```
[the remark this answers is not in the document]
Tonisal-byte: We do this on purpose to pull the distro config rather than the component specific distro config. Render does the same thing when creating its mock processor.

internal/providers/sourceproviders/sourcemanager.go
```diff
@@ -318,6 +392,19 @@ func (m *sourceManager) fetchSourceFile(
 		return nil
 	}
 
+	// Try each registered file provider. Providers return [ErrNotFound] to signal
+	// they don't handle this reference; any other error is fatal.
+	for _, provider := range m.fileProviders {
+		err := provider.GetFile(ctx, component, *fileRef, destDirPath)
+		if err == nil {
+			return nil
```
dmcilvaney, line 433: We skip hash validation here I think, in theory if the archive is deterministic we should be able to validate against the sources file right?

internal/providers/sourceproviders/customsourceprovider.go
```diff
@@ -0,0 +1,251 @@
...
+	runner.AddBindMount(scriptTmpDir, customGenScriptDir)
+	runner.AddBindMount(genOutputTmpDir, customGenOutputDir)
+
+	if err := execScriptInChroot(ctx, runner, ref); err != nil {
+		return err
+	}
+
+	// Package the output directory as a deterministic archive whose format is
+	// inferred from the filename extension (e.g., .tar.gz, .tar.xz).
+	comp, compErr := archive.DetectCompression(ref.Filename)
```
dmcilvaney: Should run this before we start mock, avoid wasted work on a typo etc.

internal/providers/sourceproviders/sourcemanager.go
```diff
@@ -260,6 +280,48 @@ func (m *sourceManager) createComponentProviders(distro ResolvedDistro) {
...
+
+	if _, statErr := env.FS().Stat(mockConfigPath); statErr != nil {
+		slog.Warn("Mock config not accessible; 'custom' origin source generation will be unavailable",
+			"path", mockConfigPath,
+			"error", statErr)
+
+		return
+	}
+
+	m.fileProviders = append(m.fileProviders, &customFileSourceProvider{
```
dmcilvaney, line 317: I'm thinking that the hash of the custom script should get folded into the fingerprint somehow, but we have an implicit assumption right now that the identity is a commit hash... not sure what to do here.

dmcilvaney: That said, if the output hash is validated, this is probably ok as-is.

internal/providers/sourceproviders/sourcemanager.go
Tonisal-byte: You're right adding the validation!

internal/providers/sourceproviders/customsourceprovider.go
```diff
@@ -0,0 +1,251 @@
...
+	if len(ref.MockPackages) > 0 {
+		if installErr := runner.InstallPackages(ctx, ref.MockPackages); installErr != nil {
+			return fmt.Errorf("failed to install mock packages for generating %#q:\n%w",
+				ref.Filename, installErr)
+		}
+	}
+
+	scriptChrootPath := filepath.Join(customGenScriptDir, ref.Script)
+
+	cmd, cmdErr := runner.CmdInChroot(ctx, []string{scriptChrootPath}, false /* interactive */, true /* pipeOutput */)
```
trungams: non-blocking: what does setting `pipeOutput = true` do here?

Tonisal-byte: This basically just pipes the output to stdout, and will be removed after Daniels #231 PR is merged in that way we can get verbose output from mock globally without having to select this one instance of mock has the capability to print to stdout.

trungams: Thanks. The rest LGTM

trungams (approved):

internal/providers/sourceproviders/customsourceprovider.go
```diff
@@ -0,0 +1,251 @@
...
+	cmd, cmdErr := runner.CmdInChroot(ctx, []string{scriptChrootPath}, false /* interactive */)
+	if cmdErr != nil {
+		return fmt.Errorf("failed to create chroot command for generating %#q:\n%w",
+			ref.Filename, cmdErr)
+	}
+
+	if runErr := cmd.Run(ctx); runErr != nil {
+		return fmt.Errorf("generation script %#q failed for source %#q:\n%w",
+			ref.Origin.Script, ref.Filename, runErr)
+	}
```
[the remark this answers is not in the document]
Tonisal-byte: If we need to retain logs from source generation then we run azldev with --verbose to print out the stdout/stderr of the mock chroot

dmcilvaney (approved):

PR #268, opened 2026-07-07
refactor(overlays): Rework OverlayMetadata to use typed URLRef list commits and enum upstream-status

liunan-ms:
## Summary

Rework `OverlayMetadata` to sharpen the review signal for AZL overlays and to make room for future per-source metadata without breaking the on-disk schema.

## Changes

- **Category rename**: `OverlayCategoryBackportDistGit` → `OverlayCategoryUpstreamBackport` (TOML value `backport-dist-git` → `upstream-backport`). Widens the category to cover Fedora dist-git *and* the component's native OSS project.
- **Typed URL references**: replace `BugRef` with a reusable `URLRef` struct. `Commits` is now `[]URLRef` (was `[]string`); `Bugs` is `[]URLRef`. The struct form leaves room for source-specific fields (tracker IDs, commit SHAs, etc.) later.
- **Upstream status enum**: replace tri-state `Upstreamable *bool` with a required typed `OverlayUpstreamStatus` (`upstreamed`, `upstreamable`, `needs-upstream-hook`, `inapplicable`, `unknown` = not yet assessed).
- **Docs / schema / snapshots**: `docs/user/reference/config/overlays.md` regenerated with the new field names, enum values, and TOML examples; `schemas/azldev.schema.json` and scenario snapshots refreshed.


internal/projectconfig/overlay_metadata.go
```diff
@@ -73,12 +73,54 @@ func (c OverlayCategory) IsValid() bool {
...
+	OverlayUpstreamStatusUpstreamed,
+	OverlayUpstreamStatusUpstreamable,
+	OverlayUpstreamStatusNeedsUpstreamHook,
+	OverlayUpstreamStatusInapplicable,
+}
+
+// IsValid reports whether s is one of the recognized [OverlayUpstreamStatus] values.
+// The empty string is not considered valid here; callers that allow the "not assessed"
+// state should test for it explicitly before calling IsValid.
+func (s OverlayUpstreamStatus) IsValid() bool {
```
trungams, line 114: Shouldn't we just require this field to be explicit instead? That would simplify the validity check as well

docs/user/reference/config/overlays.md
```diff
@@ -129,6 +129,32 @@ bugs = [
...
 
+### Upstream status
+
+`upstream-status` classifies the overlay's relationship to its upstream project. It answers "why are we carrying this?" and "what would it take to drop it?" Reviewers use it to decide whether to push back on a change or ask for an upstream link.
+
+| Value | Meaning |
+|-------|---------|
+| `upstreamed` | The change is already in Fedora / the upstream project. The overlay is carried only until AZL bumps past the fix. This is the implied default for the `upstream-backport` category. |
+| `upstreamable` | The change could be sent upstream as-is. Reviewers should ask the author to link the upstream PR. |
+| `needs-upstream-hook` | An AZL specialization that today requires invasive spec edits; upstream could add a `bcond`, `%if`, or config knob that would let us drop the overlay. Reviewers should ask whether the hook can be upstreamed instead of the change itself. |
```
trungams, line 140: I'm not sure I understand the difference between `upstreamable` vs `needs-upstream-hook`. The way I'm reading it - if an overlay carries either of these status, then it means the change can be sent upstream for reviews. In which situation should one use the latter tag/status?

liunan-ms: Good call, the doc should make the distinction sharper. The difference is what would go upstream:

**upstreamable**: the patch we're carrying is itself upstream-shaped. Sending the exact same diff (or something very close) to Fedora / the upstream project would plausibly be accepted.
**needs-upstream-hook**: the change is AZL-specific and would not be accepted upstream as-is (e.g. hard-coding `Vendor: Microsoft`, disabling a feature only we don't want, swapping a default path). But upstream could add a `bcond / %if / config` knob that would let us express the same customization without patching the spec. 

So both mean "there's an upstream action available," but `upstreamable` upstreams the change and `needs-upstream-hook` upstreams a mechanism that makes the change unnecessary. I'll tighten the table wording to say this explicitly.

internal/projectconfig/overlay_metadata.go
liunan-ms: I think you're right and here's no harm to make it required, plus it'll push the overlay contributor to pick a definite status for the change.

internal/projectconfig/overlay_metadata.go
```diff
@@ -87,19 +130,21 @@ type BugRef struct {
...
+	Category OverlayCategory `toml:"category" json:"category" jsonschema:"required,enum=upstream-backport,enum=azl-pruning,enum=azl-compatibility,enum=azl-dep-missing-workaround,enum=azl-branding-policy,enum=azl-disable-flaky-tests,enum=azl-disable-unsupported-tests,enum=azl-security-compliance,enum=azl-release-management,enum=azl-platform-adaptation,title=Category,description=Classification of the overlay's intent"`
 
-	// Commits lists URLs of upstream commits (typically Fedora dist-git or upstream-project
-	// commits) that this overlay backports or references.
-	Commits []string `toml:"commits,omitempty" json:"commits,omitempty" validate:"omitempty,dive,http_url" jsonschema:"title=Commits,description=URLs of upstream commits this overlay backports or references"`
+	// Commits references upstream commits (typically Fedora dist-git or upstream-project
+	// commits) that this overlay backports or references. Each entry must carry an
+	// http(s) URL (see [URLRef]).
+	Commits []URLRef `toml:"commits,omitempty" json:"commits,omitempty" validate:"omitempty,dive" jsonschema:"title=Commits,description=Upstream commits this overlay backports or references"`
```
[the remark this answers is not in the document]
liunan-ms: The previous version is not used by any project, so it'll not break anything.

docs/user/reference/config/overlays.md
```diff
@@ -196,8 +233,9 @@ Per-overlay `metadata` is **not allowed** inside an overlay file — the file-le
 ```toml
 # overlays/0001-cve-2024-1234.overlay.toml
 [metadata]
-category = "backport-dist-git"
-commits = ["https://src.fedoraproject.org/rpms/mypackage/c/abc123def456"]
+category = "upstream-backport"
+upstream-status = "upstreamed"
```
Tonisal-byte, line 237: Question, if the current status of the overlay is 'upstreamed' does this mean that the latest upstream ver/rel contains said change but we are deciding to stay on the current ver/rel for other factors? Will we go back and update the metadata for an overlay if it does become upstreamed but we decide not to move the input-commit?

docs/user/reference/config/overlays.md
```diff
@@ -129,6 +129,42 @@ bugs = [
...
+| `inapplicable` | A permanent AZL-only deviation. Reviewers should push back on why we have to fork upstream forever. |
+| `unknown` | The author has not yet assessed the overlay's upstream story. Reviewers should push for a definite status before approving. |
+
+`upstream-status` is required whenever `[metadata]` is present. On an `upstream-backport` overlay only `upstreamed` and `upstreamable` are allowed — any other value (`needs-upstream-hook`, `inapplicable`, `unknown`) is a validation error.
+
+#### `upstreamable` vs. `needs-upstream-hook`
+
+Both statuses mean "there is an upstream action available," but they differ in *what* would go upstream:
+
+- **`upstreamable`** — the patch we're carrying is itself upstream-shaped. Sending the same (or a close-to-identical) diff to Fedora / the upstream project would plausibly be accepted. The overlay contributor should open the upstream PR and link it, so we can drop this once it lands."
```
Tonisal-byte: nit: I think this has a trailing `"`

Tonisal-byte (approved):
Looks good overall, approved!

docs/user/reference/config/overlays.md
liunan-ms: fixed

docs/user/reference/config/overlays.md
liunan-ms: Yes, `upstreamed` means the fix is already available in fedora (its main branch / current dist-git tip / released tarball) , but we haven't bumped our pin yet. 

Once we bump past the fix, the overlay becomes redundant and is deleted, the metadata doesn't need to be "updated," instead the whole overlay should go away.

If we choose never to bump past the fix, the overlay stays; nothing about the metadata changes because it's still accurate (the change is upstream and we're just carrying our own copy).

docs/user/reference/config/overlays.md
```diff
@@ -129,6 +129,42 @@ bugs = [
...
+
+| Value | Meaning |
+|-------|---------|
+| `upstreamed` | The change is already in Fedora / the upstream project. The overlay is carried only until AZL bumps past the fix. |
+| `upstreamable` | The change could be sent upstream as-is. Reviewers should ask the author to link the upstream PR. |
+| `needs-upstream-hook` | An AZL specialization that today requires invasive spec edits; upstream could add a `bcond`, `%if`, or config knob that would let us drop the overlay. Reviewers should ask whether the hook can be upstreamed instead of the change itself. |
+| `inapplicable` | A permanent AZL-only deviation. Reviewers should push back on why we have to fork upstream forever. |
+| `unknown` | The author has not yet assessed the overlay's upstream story. Reviewers should push for a definite status before approving. |
+
+`upstream-status` is required whenever `[metadata]` is present. On an `upstream-backport` overlay only `upstreamed` and `upstreamable` are allowed — any other value (`needs-upstream-hook`, `inapplicable`, `unknown`) is a validation error.
```
trungams, line 144: This is outside the scope of this PR, but do we allow metadata to be empty/unspecified?

liunan-ms: Yes, we allow metadata unspecified for migration as we have to incrementally annotate the existing overlays but metadata cannot be empty as category and upstream-status are required fields for it.

trungams: ah okay. That makes sense

trungams (approved):

PR #269, opened 2026-07-07
feat(overlay-files): support {component} placeholder for pattern-discovered components

liunan-ms:
## Summary

Adds a `{component}` placeholder for project-level `overlay-files` entries. The resolver globs matching paths, extracts each captured segment as a component name, and synthesizes components not otherwise declared. Explicit component declarations shadow pattern-discovered matches.

## Details

- **projectconfig**: validates the placeholder per scope. Non-project scopes (distro version, component group, per-component) reject the placeholder. Project scope accepts it but rejects `\` as a path separator and glob metachars (`*?[`) *before* the placeholder (permissive in the suffix).
- **components**: new `pattern_discovery.go` walks project-level `overlay-files` entries containing `{component}`, resolves matching filesystem paths via doublestar, and captures the component name from the placeholder position. `Resolver` caches discovery results per-instance so repeated calls don't re-glob.
- **resolver**: `FindAllComponents` unions the three sources — group members, loose declared components, pattern-discovered components — and emits sorted output for determinism.
- **schema / docs**: JSON schema regenerated; user-facing `docs/user/reference/config/overlays.md` documents the placeholder.

### DIFF
--- a/internal/app/azldev/core/components/pattern_discovery.go
+++ b/internal/app/azldev/core/components/pattern_discovery.go
@@ -1,0 +1,217 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package components
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"path"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/bmatcuk/doublestar/v4"
+	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev"
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileutils"
+)
+
+// patternDiscoveredComponent captures the result of expanding a single
+// project-level [projectconfig.ComponentConfig.OverlayFiles] entry that
+// contains the `{component}` placeholder.
+type patternDiscoveredComponent struct {
+	// Name of the discovered component (the captured {component} segment).
+	Name string
+	// ReferenceDir is the absolute directory the pattern captured for this
+	// component (i.e. the prefix directory + captured name). Used as the
+	// component's reference directory for downstream operations.
+	ReferenceDir string
+	// OverlayFiles is the deduplicated, sorted list of absolute overlay file
+	// paths matched by the winning pattern for this component.
+	OverlayFiles []string
+	// SourcePattern is the exact entry string that produced this match, used
+	// in shadow-lint diagnostics.
+	SourcePattern string
+}
+
+// discoverComponentsFromPatterns walks the project-level
+// default-component-config, expands every 'overlay-files' entry that contains
+// the `{component}` placeholder, and returns the resulting pattern-discovered
+// components. Two entries at project scope producing the same component name
+// is a hard error.
+//
+// The caller is responsible for further precedence against explicitly-declared
+// components (that override happens at the resolver layer; see
+// [Resolver.patternDiscoveredComponents]).
+func discoverComponentsFromPatterns(env *azldev.Env) (map[string]*patternDiscoveredComponent, error) {
+	result := make(map[string]*patternDiscoveredComponent)
+
+	config := env.Config()
+	if config == nil {
+		return result, nil
+	}
+
+	var patterns []string
+
+	for _, entry := range config.DefaultComponentConfig.OverlayFiles {
+		if strings.Contains(entry, projectconfig.OverlayFilesComponentPlaceholder) {
+			patterns = append(patterns, entry)
+		}
+	}
+
+	if len(patterns) == 0 {
+		return result, nil
+	}
+
+	for _, pattern := range patterns {
+		byName, err := expandSinglePattern(env, pattern)
+		if err != nil {
+			return nil, fmt.Errorf("project 'overlay-files': %w", err)
+		}
+
+		for name, match := range byName {
+			if existing, ok := result[name]; ok {
+				return nil, fmt.Errorf(
+					"%w: component %#q is discovered by both %#q and %#q; "+
+						"restrict one of the patterns or rename one of the directories",
+					ErrPatternDiscoveryCollision, name, existing.SourcePattern, match.SourcePattern,
+				)
+			}
+
+			result[name] = match
+		}
+	}
+
+	return result, nil
+}
+
+// expandSinglePattern expands a single validated pattern and groups matched
+// files by captured component name.
+func expandSinglePattern(env *azldev.Env, pattern string) (map[string]*patternDiscoveredComponent, error) {
+	prefix, suffix := projectconfig.SplitOverlayFilesPlaceholder(pattern)
+
+	// Substitute a single-segment wildcard for {component} to build the actual
+	// glob. Because {component} is validated as a whole path segment (see
+	// ConfigFile.Validate), replacing it with "*" produces a well-formed
+	// doublestar pattern that captures exactly one path segment.
+	globPattern := prefix + "*" + suffix
+
+	// doublestar.WithFailOnIOErrors() is intentionally not set, mirroring
+	// findComponentGroupSpecPaths: we may legitimately hit unreadable subtrees
+	// under the project root. Non-IO errors (bad pattern, etc.) are still
+	// returned.
+	matches, err := fileutils.Glob(env.FS(), globPattern, doublestar.WithFilesOnly())
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand pattern %#q:\n%w", pattern, err)
+	}
+
+	byName := make(map[string]*patternDiscoveredComponent)
+
+	for _, match := range matches {
+		name, referenceDir, err := extractCapturedName(pattern, prefix, match)
+		if err != nil {
+			// Skip matches we can't parse (should not happen with a validated
+			// pattern, but guard anyway).
+			slog.Debug("pattern-discovery: skipping unparseable match",
+				"pattern", pattern, "match", match, "err", err)
+
+			continue
+		}
+
+		// Reject matches whose captured segment is not a safe component name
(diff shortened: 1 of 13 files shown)


internal/projectconfig/overlay_files_placeholder.go
```diff
@@ -0,0 +1,140 @@
...
+// pattern-discovery extracts the captured component name by literal-prefix
+// matching against the concrete glob result. Metacharacters after the
+// placeholder are allowed. Returns nil when the entry is well-formed.
+func validateOverlayFilesPlaceholder(entry string) error {
+	if entry == "" {
+		return fmt.Errorf("%w: entry is empty", ErrInvalidOverlayFilesEntry)
+	}
+
+	count := strings.Count(entry, OverlayFilesComponentPlaceholder)
+	if count == 0 {
```
dmcilvaney, line 45: I think this stops us having non-component placeholder entries, ie have `overlay-files = ["static/overlays/*.toml"]`. If this is expected, docs need to be updated (overlays.md:166) to match.

internal/projectconfig/overlay_files_placeholder_test.go
```diff
@@ -0,0 +1,114 @@
...
+		{name: "missing placeholder", entry: "comps/*/overlays/*.overlay.toml", wantErr: true},
+		{name: "multiple placeholders", entry: "comps/{component}/{component}/x.toml", wantErr: true},
+		{name: "placeholder is prefix substring", entry: "prefix-{component}/x.toml", wantErr: true},
+		{name: "placeholder is suffix substring", entry: "{component}-suffix/x.toml", wantErr: true},
+		{name: "backslash separator rejected", entry: "comps\\{component}\\x.toml", wantErr: true},
+		{name: "glob metachar before placeholder", entry: "**/{component}/*.overlay.toml", wantErr: true},
+		{name: "star before placeholder", entry: "comps*/{component}/*.overlay.toml", wantErr: true},
+		{name: "whole segment at start", entry: "{component}/overlays/*.overlay.toml", wantErr: false},
+		{name: "whole segment in middle", entry: "comps/{component}/overlays/*.overlay.toml", wantErr: false},
+		{name: "whole segment at end", entry: "comps/overlays/{component}", wantErr: false},
```
dmcilvaney: I think this might be an error, won't it glob files with the name of the component, not directories? I think it probably always needs to be afolder right?

internal/app/azldev/core/components/pattern_discovery.go
```diff
@@ -0,0 +1,229 @@
...
+	referenceDir = path.Clean(prefix + name)
+
+	return name, referenceDir, nil
+}
+
+// logShadowedByDeclaration emits a warning when an explicit component
+// declaration overrides pattern-discovered overlay files for the same name.
+func logShadowedByDeclaration(shadowed *patternDiscoveredComponent) {
+	slog.Warn(
+		"project 'overlay-files' pattern match shadowed by explicit component declaration; overlay files will not be applied",
```
dmcilvaney, line 208: I think this might miss shadowing from project defaults by groups. Not sure if thats something worth an error though.

internal/app/azldev/core/components/resolver.go
```diff
@@ -167,6 +179,40 @@ func (r *Resolver) FindAllComponents() (components *ComponentSet, err error) {
...
+
+	for _, name := range slices.Sorted(maps.Keys(discovered)) {
+		entry := discovered[name]
+
+		if components.Contains(name) {
+			// Only warn when the explicit declaration sets its own overlay-files,
+			// which suppresses inheritance (see [projectconfig.ComponentConfig.MergeUpdatesFrom]).
+			// A declaration with no overlay-files inherits the pattern and the
+			// discovered files still apply, so there's nothing to warn about.
+			if declared, ok := r.env.Config().Components[name]; ok && declared.OverlayFiles != nil {
```
dmcilvaney: Might want to use `len(declared.OverlayFiles) > 0` rather than nil, if I recall from the previous PR the way to disable inheritance is to set `overlay-files=[]`. 50/50 on this, we could still warn because thats weird, but it is intentionally explicitly saying "no overlays" for a component.

internal/projectconfig/overlay_files_placeholder.go
```diff
@@ -0,0 +1,140 @@
...
+	tail := idx + len(OverlayFilesComponentPlaceholder)
+	if tail < len(entry) && entry[tail] != '/' {
+		return fmt.Errorf(
+			"%w: %q in entry %#q must be a whole path segment (followed by %q or end of entry)",
+			ErrInvalidOverlayFilesEntry, OverlayFilesComponentPlaceholder, entry, "/",
+		)
+	}
+
+	prefix := entry[:idx]
+	if globMetaIdx := strings.IndexAny(prefix, "*?["); globMetaIdx >= 0 {
```
dmcilvaney, line 98: Might be able to add https://pkg.go.dev/github.com/bmatcuk/doublestar/v4#ValidatePattern here as well, get early signal back on a bad glob pattern.

internal/app/azldev/core/components/pattern_discovery.go
```diff
@@ -0,0 +1,229 @@
...
+// discoveredComponentsCache holds a lazy result of
+// [discoverComponentsFromPatterns] so the resolver can call it repeatedly
+// without re-globbing the filesystem.
+type discoveredComponentsCache struct {
+	once    sync.Once
+	entries map[string]*patternDiscoveredComponent
+	err     error
+}
+
+func (c *discoveredComponentsCache) load(env *azldev.Env) (map[string]*patternDiscoveredComponent, error) {
```
dmcilvaney: Found this a while back, might do the trick here: https://pkg.go.dev/sync#OnceValues

internal/app/azldev/core/components/pattern_discovery.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Ignore IO errors here, mirroring findComponentGroupSpecPaths: we may
+	// legitimately hit unreadable subtrees under the project root.
+	matches, err := fileutils.Glob(env.FS(), globPattern, doublestar.WithFilesOnly())
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand pattern %#q:\n%w", pattern, err)
+	}
+
+	byName := make(map[string]*patternDiscoveredComponent)
+	// Deduplicate files per component before appending to preserve stable ordering.
```
dmcilvaney: Curious if this is needed, my understanding was that glob would deduplicate internally.

liunan-ms:
Thanks @dmcilvaney for the review! After thinking about the scope and usage of `overlay-files` in azurelinux, I think it'd be better to restrict `overlay-files` scope to `project + component` only, a few reasons:

- All components now live in `base/comps/` with per-component overlays, so a single project-level `{component}` pattern is enough for every existing case. Broader scopes (distro/group) are out-of-scope for this PR and can be added later if a real requirement appears.
- Simpler validation and no risk of users accidentally declaring `overlay-files` at multiple levels and having to reason about merge/override precedence.

I update the implementation and now the scope rule is enforced. `overlay-files` allowed in:

- project `default-component-config`: must contain` {component}` (pattern-discovery + inheritance)
- `[components.X]`: plain globs relative to the declaring `.comp.toml`
- Reject (with clear errors) in _distro-version_ and _component-group_ `default-component-config`.

internal/projectconfig/overlay_files_placeholder.go
liunan-ms: This is intentional. Project-level entries without `{component}` are rejected because they would silently apply the same files to every component. Updated the doc.

internal/projectconfig/overlay_files_placeholder_test.go
liunan-ms: Great catch, updated the check and test.

internal/app/azldev/core/components/pattern_discovery.go
liunan-ms: This would be resolved by the scope restriction, now rejecting overlay-files in distro-version and component-group default-component-config entirely.

internal/app/azldev/core/components/resolver.go
liunan-ms: Fixed. Changed to `len(declared.OverlayFiles) > 0`.

internal/projectconfig/overlay_files_placeholder.go
liunan-ms: Now using `doublestar.ValidatePattern` on the substituted glob validation.

internal/app/azldev/core/components/pattern_discovery.go
liunan-ms: switched to using `sync.OnceValues`

internal/app/azldev/core/components/pattern_discovery.go
liunan-ms: `doublestar.Glob` returns unique matches per call so the seen map was dead code and has been removed.

internal/app/azldev/core/components/pattern_discovery.go
```diff
@@ -0,0 +1,204 @@
...
+// [Resolver.patternDiscoveredComponents]).
+func discoverComponentsFromPatterns(env *azldev.Env) (map[string]*patternDiscoveredComponent, error) {
+	result := make(map[string]*patternDiscoveredComponent)
+
+	config := env.Config()
+	if config == nil {
+		return result, nil
+	}
+
+	var patterns []string
```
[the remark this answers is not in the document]
liunan-ms: This is a false alert. `env.Config()` returns `*ProjectConfig`, and on `ProjectConfig`, `DefaultComponentConfig` is a value type (`ComponentConfig`, not `*ComponentConfig`). The nil `*ConfigFile.DefaultComponentConfig` field is a different struct. There is no panic risk. Existing `config == nil` guard on lines 52-54 handles the only real nil case.

dmcilvaney (approved):

PR #271, opened 2026-07-08
feat(source-files): generative source dependencies

Tonisal-byte:
Adds support for custom source generation inputs.

Custom source-files entries can now declare `origin.inputs`, a list of already-fetched source filenames that azldev stages next to the generation script inside the mock chroot. This lets scripts consume upstream tarballs or earlier source files directly by relative filename, without knowing azldev’s internal staging paths.

Also updates source preparation ordering so component sources are fetched before custom source generation runs, making upstream source tarballs available as inputs. Custom script stdout/stderr remain quiet by default, but are streamed with `--verbose` and included in failure diagnostics.

```toml
[[components.telegraf.source-files]]
filename = "telegraf-1.38.2-vendor-stripped.tar.gz"
hash-type = "SHA512"
hash = "abc123..."
origin.type = "custom"
origin.script = "gen-telegraf-vendor-stripped.sh"
origin.inputs = ["telegraf-1.38.2.tar.gz"]
origin.mock-packages = ["tar", "gzip"]
```

Inside gen-telegraf-vendor-stripped.sh, the input is available beside the script:
```bash
#!/usr/bin/env bash
set -euo pipefail

tar -xf ./telegraf-1.38.2.tar.gz
# write final generated contents into /azldev-gen/output/
```

### DIFF
--- a/docs/user/reference/config/components.md
+++ b/docs/user/reference/config/components.md
@@ -333,13 +333,15 @@ origin    = { type = "download", uri = "https://example.com/repo/.../shimx64.efi
 
 Use `origin.type = "custom"` when a source archive must be assembled or modified (e.g. stripping sensitive test fixtures from an upstream tarball). azldev runs the script inside a fresh mock chroot — the script **must write all output to `/azldev-gen/output/`**, which azldev packages into the archive named by `filename`. Network access is always enabled; the mock config comes from the project's default distro.
 
+Custom sources are regenerated on every source preparation rather than restored from lookaside. The generated archive is validated against its configured hash, so changes to the script or its inputs fail with a hash mismatch until the hash is intentionally refreshed.
+
 The `script`, `mock-packages`, and `inputs` fields are nested under `[origin]`:
 
 | Field | TOML Key | Type | Required | Description |
 |-------|----------|------|----------|-------------|
 | Script | `origin.script` | string | **Yes** | Script filename (relative to the component's spec dir) to run in mock. Required for `origin.type = "custom"`. |
 | Mock packages | `origin.mock-packages` | array of string | No | Extra RPM packages to install in the mock chroot before the script runs. |
-| Inputs | `origin.inputs` | array of string | No | Filenames to make available in the mock chroot before the script runs. Each file must already be present in the fetched source output directory — upstream source tarballs, sidecar files (patches, scripts), and any earlier `source-files` entries are all placed there by the upstream fetch before custom scripts run. |
+| Inputs | `origin.inputs` | array of string | No | Unique filenames to make available in the mock chroot before the script runs. Each file must already be present in the fetched source output directory — upstream source tarballs, sidecar files (patches, scripts), and any earlier `source-files` entries are all placed there by the upstream fetch before custom scripts run. |
 
 On first use, omit `hash` and run `prep-sources --allow-no-hashes` to generate the archive and print its hash, then copy it into the TOML.
 
--- a/docs/user/reference/config/components.md
+++ b/docs/user/reference/config/components.md
@@ -333,12 +333,13 @@ origin    = { type = "download", uri = "https://example.com/repo/.../shimx64.efi
 
 Use `origin.type = "custom"` when a source archive must be assembled or modified (e.g. stripping sensitive test fixtures from an upstream tarball). azldev runs the script inside a fresh mock chroot — the script **must write all output to `/azldev-gen/output/`**, which azldev packages into the archive named by `filename`. Network access is always enabled; the mock config comes from the project's default distro.
 
-The `script` and `mock-packages` fields are nested under `[origin]`:
+The `script`, `mock-packages`, and `inputs` fields are nested under `[origin]`:
 
 | Field | TOML Key | Type | Required | Description |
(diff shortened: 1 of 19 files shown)


docs/user/reference/config/components.md
```diff
@@ -350,8 +351,11 @@ hash      = "abc123..."               # from: prep-sources --allow-no-hashes
 origin.type          = "custom"
 origin.script        = "gen-yara-stripped.sh"    # relative to the component's spec directory
 origin.mock-packages = ["cmake"]                 # omit if not needed
+origin.inputs        = ["yara-4.5.4.tar.gz"]     # available to the script as ./yara-4.5.4.tar.gz
 ```
 
+> **Note:** Upstream source tarballs are automatically available as inputs before running custom generation scripts. There is no need to re-declare an upstream file in `source-files` to use it as an input.
+
```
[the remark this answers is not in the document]
Tonisal-byte: If we use the 'withSkipLookaside' flag the custom source wouldn't begin its generation, thus this point is not valid.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -238,13 +227,21 @@ func (p *sourcePreparerImpl) PrepareSources(
...
 
-	// Use the source manager to fetch the component (spec file and sidecar files).
+	// Fetch the component first (spec, sidecar files, and upstream source tarballs).
 	err := p.sourceManager.FetchComponent(ctx, component, outputDir, fetchOpts...)
 	if err != nil {
 		return fmt.Errorf("failed to fetch sources for component %#q:\n%w",
 			component.GetName(), err)
 	}
 
+	// Fetch custom and downloaded source files.
```
dmcilvaney, line 237: I think there is an ordering issue here, `FetchComponent` can grab a dist-git file  first, and if it happens to have the same name as something in `source-files`, we skip further checks. We should probably just not allow things like combining `replace-upstream` with this.

internal/projectconfig/configfile.go
```diff
@@ -247,6 +249,20 @@ func validateCustomSourceRef(ref SourceFileReference, componentName string) erro
...
+					"invalid 'inputs' entry %#q for source file %#q in component %#q:\n%w",
+					input, ref.Filename, componentName, err)
+			}
+
+			if input == ref.Origin.Script {
+				return fmt.Errorf(
+					"'inputs' entry %#q for source file %#q in component %#q conflicts with 'script' filename",
+					input, ref.Filename, componentName)
+			}
+		}
```
liunan-ms, line 276: Should the validation also check if there is duplicate file names in the inputs?

internal/app/azldev/core/sources/sourceprep.go
Tonisal-byte: In the latest iteration I make sure source preparation never downloads any file already defined in `source-files`. Of course we continue to support `replace-upstream` to ensure the dev understands they are replacing a source already present in the upstream

internal/providers/sourceproviders/sourcemanager_internal_test.go
```diff
@@ -0,0 +1,125 @@
...
+	_ components.Component,
+	ref projectconfig.SourceFileReference,
+	destDirPath string,
+) error {
+	p.called = true
+
+	return fileutils.WriteFile(p.fs, filepath.Join(destDirPath, ref.Filename), nil, fileperms.PublicFile)
+}
+
+func TestFetchSourceFile_ConfiguredSourceReplacesExistingFile(t *testing.T) {
```
dmcilvaney, line 41: I think we are missing a test case for a custom script with the wrong hash (should flag).

scenario/__snapshots__/TestSnapshotsContainer_config_generate-schema_stdout_1.snap
```diff
@@ -730,6 +730,14 @@
           "type": "array",
           "title": "Mock packages",
           "description": "RPM packages to install in the mock chroot before running the generation script. Only valid when origin type is 'custom'."
+        },
```
dmcilvaney, line 738: Could probably benefit from a scenario test that shows a flow something like:
Add new custom script -> generate sources -> validate sources changed via script -> modify script -> generate sources -> ensure error on bad hash.

internal/providers/sourceproviders/sourcemanager.go
```diff
@@ -452,6 +432,32 @@ func (m *sourceManager) fetchSourceFile(
...
+	lookasideErr := m.tryLookasideDownload(ctx, httpDownloader, component, fileRef, destPath)
+	if lookasideErr == nil {
+		return true
+	}
+
+	slog.Debug("Lookaside cache download failed",
+		"filename", fileRef.Filename,
+		"error", lookasideErr)
+
+	return false
```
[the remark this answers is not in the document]
Tonisal-byte: Non-issue, conflicts like this will never happen because FetchComponent only ever downloads sources not already defined as a component `source-file` in the TOMLs.

dmcilvaney (approved):

liunan-ms (approved):

PR #275, opened 2026-07-10
feat(overlays): add archive overlay engine and config model

Tonisal-byte:
[Part 1 of 2] Adds the foundation for archive-scoped overlays. 

This introduces the archive field on component overlays, allowing supported file overlays to target files inside source archives rather than loose files in the sources tree. It adds validation, fingerprint handling, schema updates, and archive utility support for safe extract/repack behavior, including compression sniffing and strict handling of unsupported tar entry types.

Example config:
```TOML
[components.apache-commons-compress]
spec = "apache-commons-compress.spec"

[[components.apache-commons-compress.overlays]]
type = "file-remove"
archive = "commons-compress-1.27.1-src.tar.gz"
file = "src/test/resources/malware/**"
description = "Remove test fixtures flagged by the signing pipeline"
```

Follow-up PR #276

### DIFF
--- a/internal/app/azldev/core/sources/archiveoverlays.go
+++ b/internal/app/azldev/core/sources/archiveoverlays.go
@@ -1,0 +1,176 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package sources
+
+import (
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+
+	"github.com/microsoft/azure-linux-dev-tools/internal/global/opctx"
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/archive"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/rootfs"
+)
+
+// processArchive extracts an archive to a temp directory, applies all overlays,
+// and deterministically repacks it with the original compression, atomically
+// replacing the original via a temp file + rename. It returns true when the
+// archive was repacked. In dry-run mode it returns (false, nil) without touching
+// the archive on disk.
+func processArchive(
+	dryRunnable opctx.DryRunnable,
+	sourcesDirPath string,
+	archiveName string,
+	overlays []projectconfig.ComponentOverlay,
+) (repacked bool, err error) {
+	archivePath := filepath.Join(sourcesDirPath, archiveName)
+
+	if dryRunnable.DryRun() {
+		slog.Info("Dry run; would apply archive overlays",
+			"archive", archiveName, "operations", len(overlays))
+
+		return false, nil
+	}
+
+	// The [archive] package operates exclusively through OS primitives ([os.Root],
+	// os.*), so extraction must use a genuine on-disk path regardless of the injected
+	// FS implementation (which may be in-memory or otherwise non-OS-backed).
+	workDir, err := os.MkdirTemp(sourcesDirPath, ".archive-overlay-")
+	if err != nil {
+		return false, fmt.Errorf("creating temp directory:\n%w", err)
+	}
+
+	defer func() {
+		if removeErr := os.RemoveAll(workDir); removeErr != nil {
+			slog.Warn("Failed to clean up archive work directory", "error", removeErr)
+		}
+	}()
+
+	// Extract the archive; compression is inferred from the filename extension.
+	// Fail on entry types we can't repack (hardlinks, devices, ...) so they
+	// aren't silently dropped from the repacked archive.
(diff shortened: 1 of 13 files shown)


internal/app/azldev/core/sources/overlays.go
```diff
@@ -53,8 +53,8 @@ func ApplyOverlayToSources(
 		}
 	}
 
-	// Apply the non-spec file component, if any.
-	if !overlay.ModifiesNonSpecFiles() {
+	// Apply the loose-file component, if any.
+	if !overlay.ModifiesLooseFiles() {
 		return nil
```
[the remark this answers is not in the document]
Tonisal-byte: That's correct, in this PR we are focusing on the archive overlay scaffolding rather than implementation.

internal/app/azldev/core/sources/archiveoverlays.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return fmt.Errorf("detecting compression for %#q:\n%w", archiveName, err)
+	}
+
+	comp, err := archive.SniffCompressionFromFile(archivePath, extComp)
+	if err != nil {
+		return fmt.Errorf("sniffing compression for %#q:\n%w", archiveName, err)
+	}
+
+	tmpFile, err := os.CreateTemp(filepath.Dir(archivePath), "."+filepath.Base(archiveName)+".repack-*")
```
dmcilvaney, line 128: We should also grab the permission bits of the input, and ensure the modified source has the same. (probably `0600` most of the time?).

internal/projectconfig/fingerprint_test.go
```diff
@@ -113,15 +114,29 @@ func TestAllFingerprintedFieldsHaveDecision(t *testing.T) {
...
-				// No tag — included by default (the safe default).
+				// Empty name — included by default (the safe default). The only
+				// option we permit is `omitempty`, which makes hashstructure skip
+				// the field when it holds its zero value (so an unset field never
+				// perturbs the hash) while still hashing it when set. Reject any
+				// other option as a likely typo.
+				if options != "" && options != "omitempty" {
+					assert.Failf(t, "invalid fingerprint tag",
+						"field %q has unrecognised fingerprint tag option %q — "+
+							"only `omitempty` is supported on included fields", key, options)
```
dmcilvaney: afaik hashstructure doesn't support `omitempty`, we probably shouldn't set that pattern here.

internal/app/azldev/core/sources/archiveoverlays.go
Tonisal-byte: You're right! Making the appropriate fix

internal/app/azldev/core/sources/archiveoverlays.go
```diff
@@ -0,0 +1,176 @@
...
+		slog.Info("Dry run; would apply archive overlays",
+			"archive", archiveName, "operations", len(overlays))
+
+		return false, nil
+	}
+
+	// The [archive] package operates exclusively through OS primitives ([os.Root],
+	// os.*), so extraction must use a genuine on-disk path regardless of the injected
+	// FS implementation (which may be in-memory or otherwise non-OS-backed).
+	workDir, err := os.MkdirTemp(sourcesDirPath, ".archive-overlay-")
```
liunan-ms: `os.MkdirTemp(sourcesDirPath, ...)` places the work dir as a sibling of the archives inside the sources tree. Is it possible that a crash happens between `MkdirTemp` and `RemoveAll`, and the `.archive-overlay-*` dir was left, then a later overlay glob or the repack would scan it?

Tonisal-byte: Good catch, making sure we only create the temp archive in the sys tmp dir.

trungams (commented):
Just a high level question: what is the intended use case for `file-search-replace`? It looks to me like something that can be achieved via a patch file.

Tonisal-byte:
> Just a high level question: what is the intended use case for `file-search-replace`? It looks to me like something that can be achieved via a patch file.

This can be achieved through a patch file yes, but the source itself must be modified because we cannot check in source code that is perceived to have malware/zipbomb type files. Both of which cause scanning to fail

Tonisal-byte (commented):
Should resolve #281

liunan-ms (approved):

internal/projectconfig/fingerprint_test.go
```diff
@@ -115,15 +115,15 @@ func TestAllFingerprintedFieldsHaveDecision(t *testing.T) {
 
 			switch tag {
 			case "":
-				// No tag — included by default (the safe default).
+				// Empty tag — included by default (the safe default).
```
dmcilvaney: nit: can probably just revert this file now

dmcilvaney (approved):

PR #276, opened 2026-07-10
feat(source-files): wire archive overlays into prep-sources with post-overlay hash tracking

Tonisal-byte:
Wires archive-scoped overlays into the source preparation pipeline.

This applies archive overlays during prep-sources, repacks modified archives, refreshes their sources file hashes, and adds origin.type = "overlay" for recording expected post-overlay archive hashes. It also ensures overlay-origin source files are not downloaded separately and validates stale configured hashes during full source prep.

Example:
```TOML
# Declares the expected post-overlay archive hash. This must have the same
# filename as the archive targeted below and replaces its upstream `sources`
# entry during render.
[[components.example.source-files]]
filename = "example-1.0.tar.gz"
hash = "..."
hash-type = "SHA2512"
origin = { type = "overlay" }
replace-upstream = true
replace-reason = "Remove bundled vendor files not shipped by Azure Linux"

# Multiple archive overlays can use the same overlay-origin entry.
[[components.example.overlays]]
type = "file-remove"
archive = "example-1.0.tar.gz"
file = "vendor/**"
description = "Remove bundled dependencies"

```


docs/user/reference/config/components.md
```diff
@@ -356,7 +356,41 @@ origin.mock-packages = ["cmake"]                 # omit if not needed
...
 
-> **Note:** Upstream source tarballs are automatically available as inputs before running custom generation scripts. There is no need to re-declare an upstream file in `source-files` to use it as an input.
+**Note:** Upstream source tarballs are automatically available as inputs before running custom generation scripts. There is no need to re-declare an upstream file in `source-files` to use it as an input.
+
+#### `"overlay"` — record a post-overlay hash
+
+- **`hash` and `hash-type` are required** for normal use. To bootstrap a new archive overlay, omit both and run `prep-sources --allow-no-hashes` once; it computes the post-overlay hash for you.
+- **`replace-upstream = true` is required** — the archive already exists in the upstream `sources` file, and this entry replaces its hash with the post-overlay value.
+- During `prep-sources` (full run), azldev verifies that the hash it computed after repacking the archive matches the stated `hash`. A mismatch means the config is stale and must be updated.
+- During `render --check-only`, the stated hash is injected directly without downloading or repacking, allowing the check to pass deterministically.
```
dmcilvaney: I think we might want to still re-pack, this is mostly a UX optimization, not a real performance optimization (i.e. we still render the whole package even with --check-only, we just don't update the files at the end).

The point is to ensure that whatever the user configured in toml matches what we have in the specs dir, and without re-running the overlay scripts etc. I'm not sure we can make that guarantee.

docs/user/reference/config/components.md
```diff
@@ -356,7 +356,41 @@ origin.mock-packages = ["cmake"]                 # omit if not needed
 origin.inputs        = ["yara-4.5.4.tar.gz"]     # available to the script as ./yara-4.5.4.tar.gz
 ```
 
-> **Note:** Upstream source tarballs are automatically available as inputs before running custom generation scripts. There is no need to re-declare an upstream file in `source-files` to use it as an input.
+**Note:** Upstream source tarballs are automatically available as inputs before running custom generation scripts. There is no need to re-declare an upstream file in `source-files` to use it as an input.
```
dmcilvaney, line 359: Can you update the AI instruction files with this info?

internal/providers/sourceproviders/sourcemanager.go
dmcilvaney: Need to add the `IsFetched()` filter here as well, or I think the file can be skipped completely.

internal/projectconfig/component.go
```diff
@@ -119,6 +132,29 @@ type SourceFileReference struct {
 	ReplaceReason string `toml:"replace-reason,omitempty" json:"replaceReason,omitempty" jsonschema:"title=Replace reason,description=Required when 'replace-upstream' is true. Human-readable explanation for the replacement." fingerprint:"-"`
 }
 
+// ValidateArchiveOverlayOrigins verifies that each archive modified by an
+// archive-scoped overlay has a matching overlay-origin source file. The origin
+// records the expected post-overlay hash used to protect render output from drift.
+func ValidateArchiveOverlayOrigins(component ComponentConfig) error {
```
dmcilvaney: blocking: We should also validate in the other direction (ie we have a overlay-origin, but no archive overlay).

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -663,6 +740,132 @@ func (p *sourcePreparerImpl) updateSourcesFile(component components.Component, o
...
+			continue
+		}
+
+		if _, ok := rehashed[line.Entry.Filename]; !ok {
+			continue
+		}
+
+		archivePath := filepath.Join(outputDir, line.Entry.Filename)
+
+		newHash, err := fileutils.ComputeFileHash(p.fs, line.Entry.HashType, archivePath)
```
dmcilvaney: Should we be using the upstream hash algorithm? afaik we only support sha256,sha512 internally, and the comparison at 855 I think will break if the algorithms are different.

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -270,65 +276,106 @@ func (p *sourcePreparerImpl) PrepareSources(
...
-		return fmt.Errorf("failed to collect overlays for component %#q:\n%w", component.GetName(), err)
+		return nil, fmt.Errorf("failed to collect overlays for component %#q:\n%w", component.GetName(), err)
 	}
 
 	if len(allOverlays) == 0 {
-		return nil
+		return nil, nil
+	}
+
+	// Archive overlays are applied first (they modify archived source files
```
dmcilvaney, line 341: Docs still claim overlays are applied in order. This can also hide a situation where a loose file remove+add can swap out an archive we just modified without any error. Maybe don't allow loose overlays to modify archive files?

internal/app/azldev/core/sources/sourceprep_test.go
```diff
@@ -99,6 +101,202 @@ func TestPrepareSources_Success(t *testing.T) {
...
+		}
+	}
+
+	require.NotNil(t, entry, "rewritten 'sources' file should still contain an entry for %q", archiveName)
+	assert.Equal(t, fileutils.HashTypeSHA256, entry.HashType, "original hash type must be preserved")
+	assert.Equal(t, repackedHash, entry.Hash, "'sources' entry must record the repacked archive's hash")
+	assert.NotEqual(t, originalHash, entry.Hash, "'sources' entry hash must have been updated")
+}
+
+func TestPrepareSources_ArchiveOverlayRequiresOverlayOrigin(t *testing.T) {
```
dmcilvaney: Should have a matrix of tests (and maybe a full e2e scenario test?)
- overlay-origin without archive overlay fails
- archive overlay without origin fails
- matching custom/download origin for an archive fails
- overlay-origin w/ configured result matches
- stale hash fails
- algorithms are stable (ie SHA512 -> SHA512, not SHA512 -> SHA256)
- local lookaside and fedora paths both keep the original archive when overlay-origin ref is used
- --skip-sources still ensures hashes are valid

internal/app/azldev/core/sources/sourceprep.go
```diff
@@ -644,7 +701,27 @@ func (p *sourcePreparerImpl) updateSourcesFile(component components.Component, o
...
+	}
+
+	// Rehash archives that were modified by archive overlays in-place.
+	if err := p.rehashModifiedEntries(existingLines, outputDir, modifiedArchives); err != nil {
+		return err
+	}
+
+	// In full prep-sources mode, cross-check any 'overlay'-origin source-file entries against
+	// the hashes that were just computed so stale stated hashes are caught immediately.
+	if !p.skipLookaside {
```
dmcilvaney: Render always sets `skipLookaside`, my feeling is we either need to track the hashes of every input (including scripts etc.), or re-run this during render. The issue is `sources` is created by render, and afaik that is the definitive truth used by koji to get source files. We need a way to ensure that a PR has the correct values for the sources file.

Tonisal-byte: Discussed offline; we can assert the render will match the TOML driven hashes that are defined for the component as all `source-files` are required to have. This will then fail during the source preparation process of the PR check if the hash does not match with the one defined.

internal/projectconfig/component.go
```diff
@@ -119,6 +132,59 @@ type SourceFileReference struct {
...
+// overlay origins are always invalid.
+func (c *ComponentConfig) ValidateArchiveOverlays(allowMissingOrigins bool) error {
+	modifiedArchives := make(map[string]bool, len(c.Overlays))
+	for _, overlay := range c.Overlays {
+		if overlay.ModifiesArchive() {
+			modifiedArchives[overlay.Archive] = true
+		}
+	}
+
+	overlayOrigins := make(map[string]bool, len(c.SourceFiles))
```
dmcilvaney, line 147: Could this break due to group/distro level defaults? ie we add a common archive overlay to a group of similar packages, or modify some source files? Might need to run this check on the final combined list to find all the conflicts.

dmcilvaney (approved):

PR #295, opened 2026-07-22
feat(ai): add component, build, runtime skills

dmcilvaney:


internal/app/azldev/agentskill/content/remove-component.md.tmpl
```diff
@@ -0,0 +1,81 @@
...
+
+```sh
+azldev comp list -p <name> -q -O json
+```
+
+Check for **reverse dependencies** first: if other components `BuildRequires` or
+`Requires` this one, removing it breaks their builds. Search the tree for the
+component name. Identify its binary subpackage names from the rendered output under
+`{{ .RenderedSpecsDir }}/<first-char>/<name>/`, then search for those names too.
+`azldev comp query` currently supports local spec sources only, not the usual upstream
```
liunan-ms, line 26: Non-blocking: the description for `azldev comp query` is correct here but it conflicts with the command's own help which says it resolves upstream sources and parses the spec, we may need to update the command help to align with the implementation in a separately PR.

liunan-ms (approved):
Thanks for adding the skills! This PR looks good to me. Only one follow-up update we may need but out of scope for this PR.

PR #296, opened 2026-07-23
feat: allow component definitions to merge across config files

ddstreet:
Remove the restriction that required each component name to be unique across all config files. Components now support field-level merge like distros: if the same component name appears in multiple files, later files' non-empty fields override earlier ones.


internal/projectconfig/loader.go
```diff
@@ -214,18 +212,28 @@ func mergeComponentGroups(resolvedCfg *ProjectConfig, loadedCfg *ConfigFile) err
...
+		resolved := component.WithAbsolutePaths(loadedCfg.dir)
+
+		if existing, ok := resolvedCfg.Components[componentName]; ok {
+			err := existing.MergeUpdatesFrom(resolved)
+			if err != nil {
+				return fmt.Errorf("failed to merge component %#q:\n%w", componentName, err)
+			}
+
+			// Preserve the latest source config file reference.
+			existing.SourceConfigFile = loadedCfg
```
reubeno, line 235: The rest of this change looks fine to me. My main question is around the `SourceConfigFile`. 

I *think* it had originally only been used for informational purposes, but I see that's now expanded a bit to usage within synthetic git history generation and presumably the rendering process.

@Tonisal-byte  /  @dmcilvaney  -- what risk, if any, do you see of this path potentially shifting if multiple TOML files provide content that contribute to the resolved component object?

reubeno (commented):
Change generally looks righteous. One main question, may need input from Toni or Daniel to understand implications.

internal/projectconfig/loader.go
Tonisal-byte: These changes should be okay for the synthetic history generation, as long as every contributing TOML is in the same Git worktree. The only potential issue I see is that there are some `overlays` and `source-file` toml fields that may contain relative paths. Once an additional component definition is created the relative reference breaks. We'd have to go back in and normalize these to absolute paths.

ddstreet: I'm not entirely clear if I need to change something to address this feedback?

internal/projectconfig/loader.go
```diff
@@ -214,18 +212,28 @@ func mergeComponentGroups(resolvedCfg *ProjectConfig, loadedCfg *ConfigFile) err
...
-			return fmt.Errorf("%w: %s", ErrDuplicateComponents, componentName)
-		}
-
 		// Fill out fields not explicitly serialized.
 		component.Name = componentName
 		component.SourceConfigFile = loadedCfg
 
-		resolvedCfg.Components[componentName] = *(component.WithAbsolutePaths(loadedCfg.dir))
+		resolved := component.WithAbsolutePaths(loadedCfg.dir)
+
```
Tonisal-byte, line 227: Should normalize overlay-files field, suggested change:

```Go
for i := range resolved.OverlayFiles {
		resolved.OverlayFiles[i] = makeAbsolute(loadedCfg.dir, resolved.OverlayFiles[i])
}
```

ddstreet: updated

dmcilvaney: @Tonisal-byte do we need to do the same for `source-files`'s `origin.script`?

internal/projectconfig/loader.go
```diff
@@ -214,18 +212,31 @@ func mergeComponentGroups(resolvedCfg *ProjectConfig, loadedCfg *ConfigFile) err
...
 		component.SourceConfigFile = loadedCfg
 
-		resolvedCfg.Components[componentName] = *(component.WithAbsolutePaths(loadedCfg.dir))
+		resolved := component.WithAbsolutePaths(loadedCfg.dir)
+		for i := range resolved.OverlayFiles {
+			resolved.OverlayFiles[i] = makeAbsolute(loadedCfg.dir, resolved.OverlayFiles[i])
+		}
+
+		if existing, ok := resolvedCfg.Components[componentName]; ok {
+			err := existing.MergeUpdatesFrom(resolved)
```
dmcilvaney, line 229: @Tonisal-byte If I recall there is some logic to detect duplicate source customizations right? Do we need to deal with that?

dmcilvaney, line 229: Looking at `MergeUpdatesFrom`, I think it doesn't handle recursive maps well. We have append set for slices, but not for maps. I'd guess that setting stuff like `packages.some-pkg.publish.rpm-channel = "foo"` might be something we want to set separately from other package bits, and currently the `packages.some-pkg` map is last one wins. Could also just weaken the documentation.

dmcilvaney:
Also, looking into why the python linter started flagging today.

dmcilvaney:
> Also, looking into why the python linter started flagging today.

#297 should fix it.

internal/projectconfig/loader.go
Tonisal-byte: Not blocking on this, should only cause a problem in one other part of the repo for `origin = custom`. Will create a separate PR fixing that issue.

liunan-ms: Agree on this concern. It doesn't block this PR because the current Azure Linux publish-channel configuration uses default-component-config, component groups, and package groups rather than splitting the same per-component package entry across files, but it'd be better to update the document to mention it and treat recursive package-map merging as a follow-up

internal/projectconfig/loader.go
liunan-ms: `overlay-files` are now normalized and synthetic history is safe within the same Git worktree. The remaining origin = "custom" script-resolution case should be tracked in a follow-up PR. Filed an issue for it: https://github.com/microsoft/azure-linux-dev-tools/issues/306

liunan-ms (approved):

PR #301, opened 2026-07-28
fix: ensure archive dir is copied deterministically

Tonisal-byte:


internal/utils/archive/archive.go
```diff
@@ -423,6 +423,39 @@ func extractEntry(root *os.Root, header *tar.Header, tarReader io.Reader, cfg ex
...
+		}
+
+		missingPaths = append(missingPaths, current)
+	}
+
+	if err := root.MkdirAll(name, fileperms.PublicDir); err != nil {
+		return fmt.Errorf("materializing directory %#q:\n%w", name, err)
+	}
+
+	for _, path := range missingPaths {
```
dmcilvaney: issue(blocking): Symlinks can cause a collision here, and I think allow them to overwrite explicit configured modes.
ie if an archive has

`real/` w/ `0700`
`x -> real` is a symlink
`x/subdir/file` causes `x/subdir` to be created
`real/subdir/` /w `0700`

The map records `x/subdir` as 0x755, and when sorting by name we end with overwriting `real/subdir` to `0x755` even though `/real/subdir` is meant to be `0700` (due to `x/*` sorting after).

Would it be possible to just set the permissions when creating missing dirs (shallow first sort), and only record explicitly changed modes to apply later? I *think* that would fix this?

internal/utils/archive/archive_linux_test.go
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Microsoft Corporation.
```
dmcilvaney, line 1: nit: file name doesn't make a lot of sense.

dmcilvaney (approved):

PR #308, opened 2026-08-10
chore: prepare release v0.3.0

dmcilvaney:


dmcilvaney:
@copilot  can you please do a simple edit pass on the changelog to align with the existing entries?

dmcilvaney:
@copilot can you add the PR numbers into all the new entires? (leave the old ones alone though).

dmcilvaney:
@copilot+gpt-5.6-sol:max looking good! Can we make them real links? And can you add an instruction file pointing to the readme that encodes these choices? Make sure it also has a good description field explaining when to load the instructions, not just a apply to frontmatter.

Don't overcomplicate it though, simple is fine.

liunan-ms (approved):

PR #323, opened 2026-08-25
feat(lockfile): completely remove the use of any 'lock' files

ddstreet:
Warning: this PR was *entirely* generated using AI. I have not looked at any of the code that's in this PR, I have only interacted with copilot to generate and test all the contained commits.

The intention of this is to completely remove the use of any 'lock' files, and instead place each component's default upstream-commit hash (i.e. the hash for the current upstream repo's branch at the specified 'snapshot' date).

Note that I made no attempt to fix any of the changelog generation issues; my reasoning is that changelog generation currently doesn't work (at least not for many/most components), so fixing it should be done in a separate PR, not here.

Personally, if I had to review this, I would not know where to begin, as the quantity of changes is (IMHO) enormous. However, much of the changes are documentation or comments, and this codebase already is (IMHO) enormous, so this PR might not be unusual for this code; I honestly do not know.

Please feel free to provide completely unfiltered feedback to this PR.


ddstreet:
To add more context:

The 'lock' files contain these fields:

- version
- manual-bump
- import-commit
- upstream-commit
- input-fingerprint
- resolution-input-hash

The `version` field is just the version of the lock file format, and can be ignored.

The `manual-bump` field is used when performing a `no-change rebuild` (a common case of this is a mass rebuild of all packages). This PR ignores it, as that is part of a following rework of how azldev 'renders' the dist-git dir for a component.

The 'import-commit' field is used only in changelog generation, and as the PR description says, this doesn't attempt to fix changelog generation which is already partly/mostly broken; that will need a follow-up PR, and won't need the info this field provides.

The 'upstream-commit' field is moved into standard TOML files, and is managed by the azldev `refresh-upstream-commit` command, which creates or updates per-component files at `base/upstream-commits/$COMPONENT.toml` (by default, the `--upstream-commit-dir` parameter can be used to change that). The `azurelinux` repo then must include these files before it includes the standard component toml files, which will override (since it's parsed later than) any upstream-commit value provided in the generated files.

The 'input-fingerprint' is used by azldev comp changed to detect if a component has changed between commits; this is replaced by standard toml parsing of the complete object model at each commit, and a simple object model component comparison between the commits, using only the specific fields that were previously used in the generation of this hash value.

The 'resolution-input-hash' is used to determine if the inputs into an upstream commit have changed. This is no longer needed because the upstream commit hash is (and always has been, actually) the actual 'source of truth' about the upstream commit.

With all those replaced or dropped, we have to generate all the upstream commit hash values in `base/upstream-commits/*.toml` files in our `azurelinux` repo; the `refresh-upstream-commit` command should be used for that. Those files also need to be included, likely in our `base/project.toml` file, as the first toml files parsed (or at least, they have to be parsed before the `base/comps/*` toml files).

I'll add detail about the release values in the next comment.

ddstreet:
This doesn't do anything to modify the logic that updates/sets the release value in rendered dist-git dirs; that logic does need a separate PR to ensure it does everything completely correctly. Additionally, the separate PR needs to handle correct changelog generation, to fill our specific changelog needs/goals as well as matching the changelog entries with our actual component EVR values for released packages.

I have not started that work yet, as I want to get feedback on this first, since it's already up to:
```
$ git diff --shortstat main..lockfile-removal 
 124 files changed, 3875 insertions(+), 12209 deletions(-)
```

I will now start working on it (using copilot), but I don't want to proceed too far into its implementation without feedback on this PR. If this is not something we think can/should get merged, then follow-on work will either be unneeded and/or implemented completely differently.

ddstreet:
Finally, I believe we need a (textual, e.g. RFCish-format) specification for the details of the process we use to (get and) convert upstream dist-git repos into local dist-git dirs (i.e. our `specs/*` dirs that contain the dist-git content for each component). That is not the only thing that the `azldev` tool does, but that is the only thing that isn't separately specified/defined elsewhere and/or implemented by some other tool or process (e.g. `azldev` can build rpms, but so can `rpmbuild`).

I think we must have a specification that covers (at a level expected by a RFC-like specification) exactly how to fully parse all our TOML files into an object model; exactly what all parts of that object model represent (e.g. valid field names, type of value, object/field usage, etc); and exactly what the process is for using our object model to convert a dist-git repo (or a local dist-git dir, as some of our components are) into a fully-processed dist-git dir. This part is, IMHO, absolutely critical and should initially be crafted based on the current (or post-merge of this and following PRs) implementation of the `azldev` tool, but thereafter the spec should be updated for any and all changes and the `azldev` tool only updated to adhere to the specification (at least, in terms of how it handles the dist-git repo -> dist-git dir conversion). I have partially started work on the specification but it's not complete and probably should wait until after this (and maybe a following) PR is merged or at least fully reviewed.

reubeno:
Thanks for sharing this, @ddstreet.

As you call out, there are also changes needed to rendering process, `azldev component build` (for local builds), and on the `koji` side to keep the distro moving forward, NEVRs not going backwards, etc. We also don't want to bundle *all* of that into 1 giant PR.

My recommendation would be to introduce a feature flag (or more if needed for decoupled features) that guard sthe behavioral changes. For example, perhaps a `--disable-lockfile` opt-in command-line option; when specified, all lockfile manipulation and querying would be skipped and instead replaced with any alternative behavior desired. We could start adding unit tests and scenario tests that exercise the new code paths, but defer actual removal of the lockfile code paths until we're ready to "flip the switch". That should also decrease the churn in this PR.

ddstreet:
> My recommendation would be to introduce a feature flag (or more if needed for decoupled features) that guard sthe behavioral changes. For example, perhaps a `--disable-lockfile` opt-in command-line option;

thanks, great suggestion, i had copilot update the PR to put all the functionality behind a global `--without-lockfile` parameter, which defaults to false, so this PR should have no effect unless that parameter is used.

internal/app/azldev/core/sources/synthistory_upstreamcommit.go
```diff
@@ -0,0 +1,214 @@
...
+
+	configFileRelPath, err := filepath.Rel(projectRepoDir, configFileAbsPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute repo-relative config path for %#q:\n%w",
+			configFileAbsPath, err)
+	}
+
+	if config.Spec.UpstreamCommit == "" {
+		return nil, nil
+	}
```
liunan-ms, line 189: I’m a little concerned that we silently skip synthetic history here when the upstream commit is empty. Skipping makes sense for an explicitly local component, but for an upstream component in `--without-lockfile` mode, should this indicate missing or broken generated state, for example, `refresh-upstream-commit` wasn’t run or the generated config wasn’t included?

Could we skip only local components and return a useful error for upstream ones, perhaps suggesting that the user run `refresh-upstream-commit`?

internal/app/azldev/cmds/component/refreshupstreamcommit.go
```diff
@@ -0,0 +1,626 @@
...
+
+  # Refresh a single component
+  azldev component refresh-upstream-commit -p curl
+
+  # Refresh components in a group
+  azldev component refresh-upstream-commit -g core
+
+  # Write generated files to a custom directory
+  azldev component refresh-upstream-commit -a --upstream-commits-dir config/commits
+
```
liunan-ms: Should these examples include `--without-lockfile`, since `refresh-upstream-commit` isn’t registered without it?

internal/app/azldev/cmds/component/refreshupstreamcommit.go
```diff
@@ -0,0 +1,626 @@
...
+	wouldPrune, orphanErr := handleOrphanConfigs(store, allComps, options)
+	if orphanErr != nil {
+		return filterRefreshDisplayResults(results), orphanErr
+	}
+
+	if options.CheckOnly {
+		wouldPrune = excludePendingRemovals(wouldPrune, results)
+
+		return refreshCheckOnlyResult(results, wouldPrune)
+	}
```
liunan-ms, line 205: Could this accidentally delete unrelated .toml files when someone points `--upstream-commits-dir` at a shared directory? Not sure if this is a concern?

liunan-ms (commented):
One question about explicit upstream pins: how should we handle a component TOML that still contains an old upstream-commit after the project snapshot is updated?

Because the component TOML is loaded after the generated upstream-commit TOML, its old pin will override the newly generated value. It seems like refresh-upstream-commit may then keep updating the generated file while the effective config continues using the old explicit pin. Should the refresh command detect and warn/error on stale explicit pins?

ddstreet:
> One question about explicit upstream pins: how should we handle a component TOML that still contains an old upstream-commit after the project snapshot is updated?

Each explicitly configured upstream commit has to be manually (person or AI) evaluated to see if it is still needed

> 
> Because the component TOML is loaded after the generated upstream-commit TOML, its old pin will override the newly generated value. It seems like refresh-upstream-commit may then keep updating the generated file while the effective config continues using the old explicit pin.

That's exactly the way it should work.

> Should the refresh command detect and warn/error on stale explicit pins?

No. It is not an error for there to be explicitly pinned components either before or after refreshing the (non-explicit) upstream commit hashes.

The dev doing the work needs to review all the explicit pins. Adding a warning or error to the refresh command just adds unnecessary noise to the operation; the dev already knows (or _should_ know) they will also need to look at the explicitly pinned components.

internal/app/azldev/core/sources/synthistory_upstreamcommit.go
ddstreet: The upstream commit should _never_ be empty for any component that uses an upstream dist-git repo.

Additionally, all the 'synthetic history' code will need to be completely replaced to get changelogs working for all components; i ignored it in this PR. I can try to get that additional PR opened asap - I feel like adding it onto this PR would increase the size and complexity even more, and make this even harder to review...do you think it would be better to include that update in this PR? If so I can try to get it updated soon...

internal/app/azldev/cmds/component/refreshupstreamcommit.go
ddstreet: thanks, i got copilot to add a commit updating these

internal/app/azldev/cmds/component/refreshupstreamcommit.go
ddstreet: > points --upstream-commits-dir at a shared directory?

Can you clarify what your concern is here? And what you mean by 'shared directory'?

ddstreet:
rebased onto latest main

trungams (approved):
Signing off on the code changes. I 've only skimmed the docs, I can take a closer look later

internal/app/azldev/core/sources/synthistory_upstreamcommit.go
liunan-ms: Makes sense, it's fine to add the update in a separate PR.

internal/app/azldev/cmds/component/refreshupstreamcommit.go
liunan-ms: By “shared directory,” I mean for example using `--upstream-commits-dir base/config` where `base/config` contains both generated files such as `curl.toml` and unrelated hand-written files such as `distro-overrides.toml`. 
`azldev --without-lockfile component refresh-upstream-commit -a --upstream-commits-dir base/config` will remove the `distro-overrides.toml`. This can be avoided if devs carefully pass a dedicated dir like `base/upstream-commits/` which only contains the generated files.

liunan-ms:
The CI check failed because documentation generated by `mage docs` differs from the committed files.
Running `mage generate` and `mage docs` and committing the changes will resolve this failure.

ddstreet:
> Merge branch 'main' into lockfile-removal

FYI, it's better to rebase a PR on top of main, not merge main into the PR.

liunan-ms (approved):

PR #325, opened 2026-08-26
feat(image): support distro kiwi config override

binujp:
## Summary

- add `kiwi-config-override` to distro version configuration, alongside that version’s Mock configuration
- resolve the configured path relative to the TOML file declaring the distro version and pass it to KIWI with `--config`
- validate the configured path and log when the override is used
- update the schema, CLI and user documentation, agent guidance, and tests

## Override location

Each build stage declares its own override on the selected distro version. For example:

```toml
[distros.azurelinux.versions.'4.0-stage1']
kiwi-config-override = "kiwi/azl4/stage1/azurelinux-4.0-kiwi-override.yml"
```

Because the path is relative to `distro/azurelinux.distro.toml`, this resolves to `distro/kiwi/azl4/stage1/azurelinux-4.0-kiwi-override.yml`. Other stages use their corresponding `distro/kiwi/azl4/<stage-name>/<override-file>` path. azldev does not discover files by walking `distro/kiwi`; it uses the exact file declared by the selected distro version.

KIWI loads this file through `--config` after its standard configuration files, so its top-level values have highest precedence.

## Design

Azure Linux needs infrastructure-specific image-build policy such as partition mapping. Keeping each override beside the corresponding staged Mock configuration is deliberate: the distro can version and review its build-environment policy while azldev remains generic.

## Validation

- `mage fix all`
- `mage docs`
- `mage scenarioUpdate`
- `mage all`

### DIFF
--- a/docs/user/reference/config/distros.md
+++ b/docs/user/reference/config/distros.md
@@ -55,6 +55,7 @@ Each distro can define multiple versions under `[distros.<name>.versions.<versio
 | Mock config | `mock-config` | string | No | Path to the mock config file for this version (architecture-independent) |
 | Mock config (x86_64) | `mock-config-x86_64` | string | No | Path to the x86_64-specific mock config file |
 | Mock config (aarch64) | `mock-config-aarch64` | string | No | Path to the aarch64-specific mock config file |
+| KIWI config override | `kiwi-config-override` | string | No | Path to KIWI configuration overrides for image builds using this version |
 | Inputs | `inputs` | [DistroVersionInputs](#inputs) | No | Per-use-case input RPM repositories for this version |
 
 ### Default Component Config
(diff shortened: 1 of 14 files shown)


internal/utils/kiwi/kiwi.go
```diff
@@ -418,6 +440,25 @@ func (r *Runner) Build(ctx context.Context) error {
 	return nil
 }
 
+func (r *Runner) generateConfigPath() (string, error) {
+	config := []byte(kiwiKPartxConfig)
+	if r.withoutKPartxOverride {
+		config = []byte("{}\n")
+	}
+
+	configPath := r.targetDir + ".yml"
```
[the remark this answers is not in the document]
binujp: Accepted. The generated config path now cleans targetDir with filepath.Clean before appending the .yml suffix, with regression coverage for a trailing path separator.

docs/user/reference/config/distros.md
```diff
@@ -176,6 +177,7 @@ dist-git-branch = "rawhide"
 [distros.azurelinux]
 description = "Azure Linux"
 default-version = "4.0"
+kiwi-config-override = "kiwi/azurelinux-kiwi-override.yml"
```
reubeno: request(blocking): From our previous discussions, this is most similar to the way we handle `mock` configuration. Could you move this to be a direct peer of `mock-config*`? (We don't need to make it arch-specific.) I think the name of the property is fine.

reubeno (changes requested):
Changes generally seem reasonable -- one main request (see below).

Please also look at copilot's comments on the PR and assess them.

docs/user/reference/config/distros.md
binujp: Ack.

reubeno (approved):

PR #326, opened 2026-09-01
chore: use github self-repo syntax for workflows

reubeno:
New zizmor warning flagged these.

Reference: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-an-action-in-the-same-repository-as-the-workflow-at-the-running-commit-recommended


Tonisal-byte (approved):
LGTM

PR #332, opened 2026-09-02
refactor(spec): add lossless structural parser and tree API

trungams:
## Summary

RPM spec files are only loosely structured, and the existing line-oriented editor has to infer section boundaries after the fact. That becomes fragile around nested conditionals, multiline macros, and sections that begin or end inside `%if` blocks.

This PR adds the private foundation for a structural editor. It parses a spec into a lossless tree of sections, conditionals, text, and macro definitions, and provides transactional query and mutation primitives over that tree. Parsing and serializing a valid spec preserves its contents byte for byte.

Nothing selects this parser in production yet, and the public overlay configuration is unchanged.

## Motivation

[Issue #214](https://github.com/microsoft/azure-linux-dev-tools/issues/214) collects several cases where line ranges are not enough to preserve section and conditional boundaries safely. A structural representation lets later changes reason about those boundaries directly without trying to evaluate RPM conditions or expand macros.

## Changes

- Parse `%if`/`%elif`/`%else`/`%endif`, including nested and empty branches.
- Keep directive-shaped text inside multiline `%define` and `%global` bodies opaque.
- Preserve backslash continuations, brace-delimited macros, Lua bodies, shell `${...}` expressions, comments, blank lines, and original ordering.
- Serialize valid input byte-for-byte.
- Reject malformed structures instead of panicking or silently truncating them.
- Add the private `structuralSpec`, `specTree`, and section handles under their final filenames and types.
- Add transactional parse/inspect/mutate helpers and section query primitives.

## Validation

- `mage build`
- `mage unit`

The cumulative stack was mechanically rebased onto `eb9fb3f` and revalidated after the rebase.

## Known limitations

The parser deliberately does not evaluate RPM macros or conditional expressions. Sections created dynamically through macro expansion are therefore not visible to it. Public editor selection and actual overlay integration arrive in later PRs in the stack.


internal/rpm/spec/structural_tree_api.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		return true
+	})
+
+	return matches
+}
+
+// RemoveSections removes sections as one transaction.
+func (t *specTree) RemoveSections(handles []*sectionHandle) error {
```
Tonisal-byte, line 112: The parser deliberately represents the preamble as an implicit section with empty name and package. `Sections("", "")` can return that handle, and `RemoveSections()` accepts it without restriction in `structural_tree_api.go:82-128`.

The existing API explicitly rejects this operation in `edit.go:810-815`. Removing the structural preamble would discard global fields such as Name, Version, Release, and sources.

internal/rpm/spec/tree.go
```diff
@@ -0,0 +1,898 @@
...
+	longClose string
+}
+
+func (state macroState) open() bool {
+	return state.depth > 0 || state.escapedBraces > 0 || state.lua != nil
+}
+
+// percentRunOpensBracedMacro reports whether the percent run at start ends in
+// an active '%{' opener. RPM escapes percent pairs, leaving only odd runs live.
+func percentRunOpensBracedMacro(content string, start int) bool {
```
Tonisal-byte, line 639: NIT: It seems only the unit test uses this function; do we still need it? I understand if it's required for later PRs in the stack :D

trungams: It is used later on when we need to preserve macro definitions when removing sections of a spec file

internal/rpm/spec/structural_tree_api.go
trungams: Thanks for the catch!

internal/rpm/spec/structural_tree_api.go
```diff
@@ -0,0 +1,290 @@
...
+func (t *specTree) Section(name, pkg string) *sectionHandle {
+	for _, section := range t.Sections(name, pkg) {
+		return section
+	}
+
+	return nil
+}
+
+// HasSection reports whether a section with name is present for any package.
+func (t *specTree) HasSection(name string) bool {
```
Tonisal-byte, line 67: Non-blocking: `HasSection` and `Sections` compare section names exactly, while isSectionHeaderLine recognizes them case-insensitively and getSectionNameAndPackageFromHeader stores the original spelling.

internal/rpm/spec/structural_tree_api.go
```diff
@@ -0,0 +1,290 @@
...
+	if err := validateSectionRemoval(t.root, sections); err != nil {
+		return err
+	}
+
+	removeSections(t.root, sections)
+
+	return nil
+}
+
+// Name returns the section keyword. The preamble has an empty name.
```
Tonisal-byte, line 135: Non-blocking: The parser treats directive-shaped lines as structure unless they occur inside %define or %global. Consequently, valid but uncommon standalone multiline `%{lua:...}` or `%{expand:...}` blocks containing %if or section-shaped text can be rejected or misclassified.

Tonisal-byte (approved):
Some comments that might be helpful to touch upon but not required

liunan-ms (approved):

PR #333, opened 2026-09-02
refactor(spec): add selectable legacy and structural editors

trungams:
## Summary

This PR puts the existing line-oriented editor and the new structural editor behind the same public `Spec` API. The legacy implementation remains intact, while the structural implementation uses the parser and tree API introduced in the previous PR.

Editor selection happens once when a spec is opened. There are no mode checks scattered through individual operations and no automatic fallback from one editor to the other. This PR also adds a curated set of real-world spec fixtures to exercise the structural implementation. Production continues to use the legacy editor until the rollout configuration is introduced at the top of the stack.

## Motivation

[Issue #214](https://github.com/microsoft/azure-linux-dev-tools/issues/214) calls for structural awareness when editing sections and conditionals. At the same time, azldev is stabilizing and we want an escape hatch during adoption. Keeping both implementations behind one facade gives projects a controlled migration path without changing the public editing API.

## Changes

- Add a private editor interface and constructor.
- Keep `Spec` as the unchanged public facade.
- Preserve the legacy visitor and line-oriented implementation.
- Add the structural implementation for section, tag, search, patch, and changelog operations.
- Preserve current-main release and provenance behavior.
- Keep structural visitors fully structural; macro bodies remain opaque to tag traversal.
- Preserve sequential search/replace overlays that temporarily pass through an unbalanced conditional state.
- Add curated real-world fixtures and deterministic parser/edit stress coverage.

## Validation

- `mage build`
- `mage unit`

The cumulative stack was mechanically rebased onto `eb9fb3f` and revalidated. The complete rollout was also exercised against the full Azure Linux structural render corpus and the default-legacy E2E suite.

## Known limitations

Section and subpackage removal does not automatically relocate `%define` or `%global` declarations from removed content. Projects must preserve any declarations still needed by surviving content explicitly. The structural editor does not evaluate conditional expressions; when ownership cannot be established statically, removal is rejected rather than guessed. There is intentionally no automatic fallback between editors.


internal/rpm/spec/structural_edit.go
```diff
@@ -0,0 +1,1044 @@
...
+			}
+
+			line.Replace(fmt.Sprintf("%s: %s", tag, value))
+
+			updated = true
+
+			return nil
+		})
+	})
+	if err != nil {
```
liunan-ms, line 62: This seems a different behavior across the editors. Structural updates every match, while legacy only updates the first, so the same public API behaves differently depending on the selected editor. Should we keep this consistent across editors? OR document the difference?

internal/rpm/spec/structural_tree_api.go
```diff
@@ -132,33 +188,236 @@ func (t *specTree) RemoveSections(handles []*sectionHandle) error {
...
+
+	matches := tagRegex.FindStringSubmatch(line)
+	if len(matches) != reSubmatchCount {
+		return "", "", false
 	}
 
-	child := &block{Kind: textBlock, Lines: lines}
-	h.block.Children = append([]*block{child}, h.block.Children...)
+	return matches[1], matches[2], true
+}
```
liunan-ms, line 409: Some comment line like `# Patch0: disabled.patch` would be parsed as a tag. Should we skip comments before matching here?

internal/rpm/spec/structural_edit.go
```diff
@@ -0,0 +1,1044 @@
...
+		if _, isMacro := isMacroDefLine(line); isMacro {
+			macroParseState, inMacroBody = macroBodyStateAfter(line, macroState{})
+
+			continue
+		}
+
+		if isSectionHeaderLine(line) {
+			currentName, currentPackage = getSectionNameAndPackageFromHeader(line)
+			sectionFound = sectionFound || (currentName == sectionName && currentPackage == packageName)
+		} else if currentName == sectionName && currentPackage == packageName {
```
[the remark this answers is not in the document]
liunan-ms:
```
Name: example
Source0: source.tar.gz

%if 0%{?with_tools}
%package tools
Summary: Helper tools
%endif

Source1: extra.tar.gz
License: MIT
```
For this example, the `InsertTag` will append a high numbered source macro to `Source0` instead of `Source1`.

internal/rpm/spec/structural_edit.go
```diff
@@ -0,0 +1,1044 @@
...
+		}
+
+		state.advanceBefore(line)
+
+		if state.matchesFilter(filterSection, filterPkg) {
+			if newLine := compiledRegex.ReplaceAllLiteralString(line, replacement); newLine != line {
+				lines[index] = newLine
+				updated = true
+			}
+		}
```
liunan-ms, line 629: Is it possible that the replacement has newline/multiple lines like `%if 1\nmake\n%endif`? Should we split replacement containing `\n` into physical-line entries before committing them?

liunan-ms (commented):
@trungams Thanks for the great work! I left some comments/questions.

internal/rpm/spec/structural_edit.go
trungams: I think both methods update every match, but the documentation on the legacy method is not correct. We also probably need to have a full specs that cover the overlay operations first. I'm working on that next after this PR

internal/rpm/spec/structural_edit.go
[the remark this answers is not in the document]
trungams: That's intended to be conservative. If the condition evaluates to true, we'll enter a new `%package` section, and `Source1` + `License` lines will then belong to the new context. Inserting a new tag right after `Source0` is safest though it may mess up the ordering of the `SourceN` tags.

This is something that should be better defined in a spec still, so I'd prefer to keep it as is for now.

internal/rpm/spec/structural_edit.go
trungams: I'd advise against doing something like that in overlays, and one of the reasons for structural editor is to reduce the use of spec search replace API. Plus the legacy editor also suffer from a similar issue, so I'd prefer not to address it in this PR

internal/rpm/spec/structural_tree_api.go
trungams: Good catch. The exact spaced form is already rejected, but `#Patch0:` was parsed as a tag and that form is common in the Azure Linux corpus. Fixed in 4e5fde9 by explicitly rejecting comment lines before tag matching, with direct parser and cross-editor visitor coverage.

trungams: Thanks for the catch. I updated this PR

internal/rpm/spec/structural_edit.go
```diff
@@ -0,0 +1,1044 @@
...
+				Package:  packageName,
+			},
+			structuralLine: line,
+		})
+	})
+
+	lines := serializeTree(root)
+	if visitErr != nil {
+		s.rawLines = lines
+
```
liunan-ms: Could we validate the serialized mutation before assigning them to `s.rawLines` on the `visitErr` path? For example, a callback could replace `Name:` with an unmatched `%if`, then return an error on the next tag.

trungams: fixed

PR #337, opened 2026-09-03
fix(config): resolve components in config dump output

### LINKED ISSUE #264: config dump omits per-file overlays and inherited default-component-config
### Summary

`azldev config dump` presents itself as showing the *"fully resolved project configuration"*, but it stops at the file-level merge and does not run the per-component resolver. As a result, two important sources of effective configuration are invisible in the dump even though builds honor them:

1. **`overlay-files` expansion** — overlays loaded from `overlays/*.overlay.toml` (per-file overlay format, [docs](https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/overlays.md#per-file-overlay-format)) never appear in `components.<name>.overlays`.
2. **Inherited `default-component-config`** — fields injected from project/distro/component-group defaults never appear on the concrete component.

Downstream commands (`build`, `render`, `prepare-sources`, `diff-sources`) *do* run the resolver (`applyInheritedDefaultsToComponent` → `ExpandResolvedOverlayFiles`), so the artifacts they produce are correct. The gap is display-only, but it makes `config dump` misleading as an authoring/debugging aid.

### Repro

Given a component using the per-file overlay format:

```toml
# base/comps/GraphicsMagick/GraphicsMagick.comp.toml
[components.GraphicsMagick]
overlay-files = ["overlays/*.overlay.toml"]
```

```toml
# base/comps/GraphicsMagick/overlays/0001-drop-libheif-build-dependency.overlay.toml
[metadata]
category = "azl-dep-missing-workaround"

[[overlays]]
type = "spec-remove-tag"
tag = "BuildRequires"
value = "libheif-devel"
```

Then:

```
$ azldev config dump -f json | jq '.components.GraphicsMagick.overlays'
null
```

But `azldev component list -p GraphicsMagick -O json | jq '.[0].Overlays'` shows the overlay, and `azldev component render -p GraphicsMagick` correctly removes the `BuildRequires: libheif-devel` line from the rendered spec.

### Impact

- Confuses users during authoring — a config that clearly "works" appears empty in the dump, prompting incorrect "my overlays aren't being applied" diagnoses.
- Blocks the documented pattern of using `azldev config dump -q -O json | jq ...` for ad-hoc queries about effective component config (works today only for fields authored inline).
- Makes the MCP `config-dump` tool less useful for coding agents that ask "what is the effective config for component X?".

### Suggested fix

Add a flag to `config dump` that runs the per-component resolver before serialization. Some options:

- **A (preferred):** `azldev config dump --resolved` (or `--expand`) — off by default to preserve today's round-trippable output; when set, iterates components through the resolver and replaces the `components` map with resolved copies (including expanded `overlays` and merged inherited defaults).
- **B:** Always dump the resolved view; drop the round-trippability guarantee. Breaking change; probably not worth it.
- **C:** Documentation-only — clarify in `Long` help and [config-file.md](https://github.com/microsoft/azure-linux-dev-tools/blob/main/docs/user/reference/config/config-file.md) that `config dump` is pre-resolver and point users to `component list -O json` for the resolved view. Cheapest fix but doesn't help the MCP-tool case.

Happy to send a PR for option A if the team agrees on the shape.

### Related code

- Dump entry point: `internal/app/azldev/cmds/config/dump.go` — marshals raw `env.Config()`.
- Resolver used by build/render/prepare-sources: `internal/app/azldev/core/components/resolver.go` — `Resolver.FindComponents` → `getComponentFromNameAndSpecPath` → `applyInheritedDefaultsToComponent` → `ExpandResolvedOverlayFiles`.
- Feature introduced by PR #256 (per-file overlays / inherited `overlay-files`); this gap has existed since then.

liunan-ms:
## Summary

`azldev config dump` previously serialized the file-level merged config, so per-file overlays (`overlay-files`) and inherited `default-component-config` were invisible even though builds honor them (issue #264).

This change resolves each component through the resolver before serialization:

- Component entries now include inherited defaults (project / distro / group) and overlays expanded from `overlay-files`.
- Components discovered via group spec globs are included.
- Runtime-only state (`Locked`, `RenderedSpecDir`) is stripped so it doesn't leak into the JSON dump.
- The loaded config is not mutated.

The dump retains the `default-component-config` blocks alongside the resolved component entries, so it is an **inspection snapshot** and is **not** intended to be reloaded as project configuration (this is called out in the updated help and docs).

## Docs

- Updated `config-system.md`, `config-file.md`, and `repos.md` to describe the resolved snapshot and drop the round-trippability claim.
- Corrected `-O json` → `-f json` in examples (`config dump` uses its own `-f/--format` flag; `-O` never controlled its output).
- Regenerated `azldev_config_dump.md` via `mage docs`.

## Testing

- `mage all` (unit + scenario + checks) passes.
- New unit tests cover inherited defaults across all layers, overlay-file expansion, runtime-field exclusion, source-config immutability, and spec-discovered components.

Fixes #264

### DIFF
--- a/internal/app/azldev/cmds/config/dump.go
+++ b/internal/app/azldev/cmds/config/dump.go
@@ -10,6 +10,7 @@ import (
 	"slices"
 
 	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev"
+	"github.com/microsoft/azure-linux-dev-tools/internal/app/azldev/core/components"
 	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
 	"github.com/pelletier/go-toml/v2"
 	"github.com/spf13/cobra"
@@ -63,9 +64,11 @@ func newDumpCmd() *cobra.Command {
 		Long: `Dump the fully resolved project configuration.
 
 Shows the merged result of all config files (embedded defaults, project
-config, includes, and any extra --config-file arguments) after inheritance
-and merge rules have been applied. Useful for debugging configuration
-issues or inspecting effective values.`,
+config, includes, and any extra --config-file arguments). Component entries
+include inherited defaults and overlays expanded from 'overlay-files'.
+
+The output is an inspection snapshot for debugging and scripting, not project
+configuration intended to be loaded again.`,
 		Example: `  # Dump config as TOML (default)
   azldev config dump
 
@@ -100,7 +103,10 @@ issues or inspecting effective values.`,
 }
 
 func DumpConfig(env *azldev.Env, format configDumpFormat) (string, error) {
-	config := portableConfigCopy(env.Config())
+	config, err := resolvedConfigCopy(env)
+	if err != nil {
+		return "", err
+	}
 
 	switch format {
 	case ConfigDumpFormatTOML:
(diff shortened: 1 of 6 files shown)


internal/app/azldev/cmds/config/dump.go
```diff
@@ -122,6 +128,26 @@ func DumpConfig(env *azldev.Env, format configDumpFormat) (string, error) {
...
+		return nil, fmt.Errorf("failed to resolve components:\n%w", err)
+	}
+
+	config := *env.Config()
+
+	config.Components = make(map[string]projectconfig.ComponentConfig, resolved.Len())
+	for _, component := range resolved.Components() {
+		componentConfig := *component.GetConfig()
+		// Drop resolver-populated runtime state so it doesn't leak into the JSON dump.
+		componentConfig.Locked = nil
```
trungams, line 143: Do you know why we don't want to include these fields in the config dump?

liunan-ms: My guess is both `locked` and `RenderedSpecDir` are generated state or data instead of human-authored configuration so config dump excludes them.

trungams: That seems quite arbitrary to me. I think we should dump the full config, then let the users of the command filter the output themselves.

PR #339, opened 2026-09-04
feat(repo): add concise comparison modes

Tonisal-byte:
## Summary
Extends azldev repo compare with focused reporting and optional content verification for comparing Koji and PMC repository inventories.

### Features
* Adds `--missing-from-right` to list package versions present only in the left repository.
* Adds `--stat` for package-level difference counts.
* Adds `--ignore-older-added-in-right` to suppress historical versions retained by PMC when Koji contains a newer EVR.
* Adds opt-in `--compare-checksums` comparison for matching package identities.
* Reports checksum and size differences as content-different.
* Reports incompatible checksum algorithms as content-comparison-skipped.


internal/app/azldev/cmds/repo/compare.go
```diff
@@ -108,14 +125,38 @@ func runCompare(
 		return nil, fmt.Errorf("loading right repositories:\n%w", err)
 	}
 
-	reports, err := repocompare.Compare(leftPackages, rightPackages)
+	if options.MissingFromRight {
+		return compareMissingFromRight(leftPackages, rightPackages, options)
+	}
```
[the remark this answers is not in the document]
liunan-ms: Is it intentionally skip the non-shared repo in filterToSharedKinds?

internal/app/azldev/cmds/repo/compare.go
```diff
@@ -58,7 +75,7 @@ not compared.`,
 }
 
 // RunCompare loads both repository inventories and returns their package identity differences.
```
liunan-ms, line 77: Could we update the comment above now that the return type is interface{}?

Tonisal-byte: Thanks for catching the `filterToSharedKinds` mistake, that was unintentional. Originally the command was producing a lot of noise due to the lack of subrepo filtering but it's not necessary anymore. The filter is now derived from the subrepos defined in the toml.

PR #345, opened 2026-09-09
fix(overlays): allocate macro source tag to avoid collisions

### LINKED ISSUE #235: [Bug]: Macro overlay source files can collide with real sources
**Bug Description**
The macro overlay hardcodes its source to `9999` (`synthesizeMacroLoadOverlays()`), which can collide (see texlive.spec](https://raw.githubusercontent.com/microsoft/azurelinux/refs/heads/4.0/specs/t/texlive/texlive.spec)

**Reproduction Steps**
1. Move textlive to its own `*.comp.toml` file
2. Add
    ```
    [components.texlive]
    [components.texlive.build]
    defines = { azl_poc_macro = "1" }
    ```
3. Run `azldev comp render -p texlive`

**Expected behavior**
Should add the macro file in a smart way that doesn't cause errors.

**Actual behavior**
```
17:09:41 INF Finishing rendered output count=1
17:09:41 ERR Failed to finish rendering component component=texlive error="mock processing failed for `texlive`:\nspectool failed: error: source 9999 defined multiple times"
17:09:41 ERR Some components failed to render errorCount=1 cancelledCount=0
```

**Logs / Output**
```
azldev comp render -p texlive 2>&1 | tee base/build/work/scratch/texlive-render.log | tail -40
17:08:02 INF Mock processor available mockConfig=/home/damcilva/repos/temp/azurelinux/distro/mock/azl4/stage2/azurelinux-4.0-x86_64.cfg
17:08:02 INF Preparing component sources count=1
17:08:02 INF   Preparing component sources component=texlive
17:08:02 INF Getting component from git repo component=texlive upstreamComponent=texlive branch=f43 upstreamCommit=f5f486ecd487e3c34a8e747c0fc561fa1eaff9fe snapshot=2026-02-24T00:00:00-08:00
17:08:02 INF     Cloning git repo repoURL=https://src.fedoraproject.org/rpms/texlive.git
17:09:23 INF Using explicit upstream commit hash commitHash=f5f486ecd487e3c34a8e747c0fc561fa1eaff9fe
17:09:23 INF     Checking out git commit commitHash=f5f486ecd487e3c34a8e747c0fc561fa1eaff9fe
17:09:23 INF     Applying overlays component=texlive
17:09:24 INF Current fingerprint differs from HEAD lock file; adding dirty entry lockFile=locks/texlive.lock
17:09:25 INF Bumping static release component=texlive oldRelease=80%{?dist} newRelease=82%{?dist} commitCount=2
17:09:25 INF Creating synthetic commit commit=1 total=2 projectHash=5d461a2449294d42c9b47ed04ad3d5935d8ce244 upstreamCommit=f5f486ecd487e3c34a8e747c0fc561fa1eaff9fe isLast=false
17:09:25 INF Creating synthetic commit commit=2 total=2 projectHash=dirty upstreamCommit=f5f486ecd487e3c34a8e747c0fc561fa1eaff9fe isLast=true
17:09:25 INF Interleaved synthetic history complete syntheticCommits=2 totalCommits=3
17:09:25 INF Initializing mock chroot for rendering
17:09:37 INF Mock chroot ready for rendering
17:09:37 INF Batch processing components in mock chroot count=1
17:09:37 INF Processing specs in mock chroot count=1
17:09:41 INF Finishing rendered output count=1
17:09:41 ERR Failed to finish rendering component component=texlive error="mock processing failed for `texlive`:\nspectool failed: error: source 9999 defined multiple times"
17:09:41 ERR Some components failed to render errorCount=1 cancelledCount=0
╭───────────┬──────────────────────────────────────────────────────┬────────┬────────────────────────────────────────────────────────────┬─────────╮
│ COMPONENT │ OUTPUT                                               │ STATUS │ ERROR                                                      │ CHANGED │
├───────────┼──────────────────────────────────────────────────────┼────────┼────────────────────────────────────────────────────────────┼─────────┤
│ texlive   │ /home/damcilva/repos/temp/azurelinux/specs/t/texlive │ error  │ mock processing failed for `texlive`:                      │ false   │
│           │                                                      │        │ spectool failed: error: source 9999 defined multiple times │         │
╰───────────┴──────────────────────────────────────────────────────┴────────┴────────────────────────────────────────────────────────────┴─────────╯
```

**azldev version**
```
╭────────────┬──────────────────────────────────────╮
│ version    │ v0.1.1-0.20260520162735-44f81cce609a │
│ gitCommit  │ N/A                                  │
│ goVersion  │ go1.26.3                             │
│ platform   │ linux/amd64                          │
│ compiler   │ gc                                   │
│ buildDate  │ N/A                                  │
│ commitDate │ N/A                                  │
│ dirtyBuild │ false                                │
╰────────────┴──────────────────────────────────────╯
```

**Environment**
- azl3

liunan-ms:
Fixes #235

## The problem

The macro-load overlay hard-coded its source as `Source9999`. When a spec already defines that tag (e.g. texlive, whose sources go up to `Source10133`), the injected tag collided:

```
spectool failed: error: source 9999 defined multiple times
```

## The fix

Allocate the macros source tag dynamically: keep the conventional `Source9999` when that number is free (so already-rendered specs stay byte-identical), and fall back to the next number after the highest existing source tag only when `Source9999` is already taken.

## Why prefer `Source9999` instead of always `highest+1`?

azldev has historically emitted the macros file as `Source9999`, and the rendered specs committed in `microsoft/azurelinux` reflect that. Always allocating `highest+1` would change the tag number for every macros-bearing spec (e.g. a spec whose highest source is `Source1` would get `Source2`), making `render` non-idempotent against the committed trees and breaking the end-to-end idempotency test. Preferring `Source9999` keeps existing output stable while still resolving the collision for the rare specs that already use `Source9999`.

## Example: texlive

texlive's spec already uses `Source9999` (and sources up to `Source10133`):

**Before:** injected a second `Source9999: texlive.azl.macros` → collides → build fails.

**After:** `AddSourceEntry` sees `Source9999` is occupied, finds `highest = 10133`, and inserts:

```rpmspec
Source10134: texlive.azl.macros
```

For an ordinary spec where `Source9999` is free, it stays `Source9999` (unchanged from today's output).

## Notes

- Adds `Spec.AddSourceEntry` and `ParseSourceTagNumber` (rejecting negative suffixes), mirroring the existing `AddPatchEntry` / `GetHighestPatchTagNumber` logic. Bare `Source:` tags count as auto-numbered slots.
- The macros load directive and source registration are returned as separate overlays so the source registration is applied **after** user overlays, ensuring it accounts for any sources the user added before choosing a number.
- The synthesized `internal-source-add` overlay is applied at source-prep time, so it has no component fingerprint impact.


internal/rpm/spec/edit.go
```diff
@@ -631,6 +632,77 @@ func ParsePatchTagNumber(tag string) (int, bool) {
...
+// next number after the highest existing source tag. Automatically numbered bare Source tags are
+// included when determining occupancy and the highest number.
+func (s *Spec) AddSourceEntry(filename string) error {
+	const preferredSourceTagNumber = 9999
+
+	highest := -1
+	unnumberedCount := 0
+	preferredOccupied := false
+
+	err := s.VisitTags(func(tagLine *TagLine, _ *Context) error {
```
Tonisal-byte, line 664: Non-blocking(edge-case): `AddSourceEntry` checks occupied numbers but never checks whether filename is already registered. Reprocessing a spec containing `Source9999: macros.azl.macros` therefore adds the same file as Source10000, and subsequent passes keep adding entries. rpmspec retains these duplicates, contradicting the stated idempotency guarantee

Tonisal-byte (approved):

trungams:
> Always allocating highest+1 would change the tag number for every macros-bearing spec (e.g. a spec whose highest source is Source1 would get Source2), making render non-idempotent against the committed trees and breaking the end-to-end idempotency test

Correct me if I'm wrong, but this only happens if we apply overlays against rendered specs, which we don't do. Since we apply overlays against upstream spec, we'd get the same `highest + 1` every time provided that the usptream spec is unchanged. So it doesn't make sense to keep the hard-coded `Source9999` tag

liunan-ms:
> > Always allocating highest+1 would change the tag number for every macros-bearing spec (e.g. a spec whose highest source is Source1 would get Source2), making render non-idempotent against the committed trees and breaking the end-to-end idempotency test
> 
> Correct me if I'm wrong, but this only happens if we apply overlays against rendered specs, which we don't do. Since we apply overlays against upstream spec, we'd get the same `highest + 1` every time provided that the usptream spec is unchanged. So it doesn't make sense to keep the hard-coded `Source9999` tag

I think you’re right. `highest + 1` would be deterministic since we always start from the upstream spec (I used it for my first version of fix and then I realized it will need a mass update in the existing rendered specs in azl repo). So keeping Source9999 is only for avoiding churn in existing rendered specs. The choice is between minimizing generated-output churn by retaining Source9999, or adopting the cleaner highest + 1 rule. I’m leaning toward the less-churn workaround here, also wondering your thoughts.

trungams:
Personally I'd prefer keeping the tool clean. The migration from using `Source9999` to the new logic in azurelinux would only be a one time change.

PR #346, opened 2026-09-09
feat(image): support per-image architecture overrides for aarch64 builds

binujp:
## Summary
Add an optional `architectures` field to image objects to specify which
architectures that image can build on.

All images need not build on all architectures. E.g. gen1 images do not build
for aarch64. We have to specify this metadata about which images build on which
architectures in an image-definition- and build-mechanism-independent manner.
`images.toml` lets us encode this metadata per image, and azldev generates the
JSON schema for consumption.

The `architectures` field is optional: an image with no declared `Architectures`
is treated as unrestricted (all recognized architectures) rather than rejected
during validation. This lets older `images.toml` files work with the newer
azldev version and gives us a window to switch over.

## Changes
- `internal/projectconfig/image.go`, `project.go`, `loader.go`: add/validate optional per-image architecture overrides
- `internal/app/azldev/cmds/image/build.go`, `list.go`: apply overrides during build/list
- `docs/user/reference/config/images.md`, `internal/app/azldev/agentskill/content/image.md.tmpl`: document the new config
- `schemas/azldev.schema.json`, scenario snapshots: regenerated via `mage docs` / `mage scenarioUpdate`

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


docs/user/reference/config/images.md
```diff
@@ -11,6 +11,13 @@ The `[images]` section defines system images (VMs, containers, etc.) that azldev
 | Capabilities | `capabilities` | [ImageCapabilities](#image-capabilities) | No | Describes features and properties of this image |
 | Tests | `tests` | [ImageTests](#image-tests) | No | Test configuration for this image |
 | Publish | `publish` | [ImagePublish](#image-publish) | No | Publishing settings for this image |
+| Architectures | `architectures` | string array | No | Architectures supported by this image |
+
+The current supported architectures are `x86_64` and `aarch64`. `architectures` is
```
tobiasb-ms, line 16: question(non-blocking): Is this the appropriate spot for this description of `Architectures`? I definitely like having it but I don't see anything else doing being described here. I wonder if there's a better place for it, but this is definitely non-blocking.

internal/app/azldev/cmds/image/list.go
```diff
@@ -135,10 +143,15 @@ func ListImages(env *azldev.Env, options *ListImageOptions) ([]ImageListResult,
...
-			Tests:               imageConfig.Tests,
-			TestsSummary:        strings.Join(imageConfig.TestNames(), ", "),
-			Publish:             imageConfig.Publish,
-			PublishSummary:      strings.Join(imageConfig.Publish.Channels, ", "),
+			Architectures:       imageConfig.Architectures,
+			ArchitecturesSummary: strings.Join(
+				imageConfig.Architectures,
+				", ",
+			),
+			Tests:          imageConfig.Tests,
```
tobiasb-ms, line 151: nit(non-blocking): I notice that you changed the indentation/justification for these fields. Was that intentional? If so, why?

docs/user/reference/config/images.md
binujp: I see your point and am also stuck where else the blurb would fit better. This field is more nuanced than the others and needs the description.

internal/app/azldev/cmds/image/list.go
binujp: This is all gofmt; one of those go ecosystem design decision which I love. No one gets to entertain opinions without having to fight the system 😃

docs/user/reference/config/images.md
liunan-ms: NIT: keeping Architectures in the top-level image config table make sense, ideally move it up next to Definition, because it is a direct property of [images.<name>], not a subtable.

binujp: I see the argument; architecture is more about what has to be done to the image vs what has to be done with the image like with publish.

liunan-ms (approved):

PR #349, opened 2026-09-09
feat(sources): use rpmdev-bumpspec for release updates

trungams:
## Summary

Reframe rpmdev-bumpspec as a temporary CLI-only preview rather than the default release path.

- Restore baseline legacy static Release calculation by default, including integer, `integer%{?dist}`, and `integer%{dist}` forms.
- Add `--rpmdev-bumpspec` only to `component build`, `component render`, and `component prepare-sources` to select the existing transactional runner.
- Reject `--rpmdev-bumpspec` with `--without-git`, and with `prepare-sources --skip-overlays`, before source preparation or runtime discovery.
- Document the preview, deferred Azure Linux migration, legacy non-standard Release error, unchanged `%autorelease`, and opt-in-only host tool requirements.

## Validation

`mage fix all`, `mage unit`, `mage check all`, `mage docs`, and final `mage all` passed. `mage all` completed scenario validation. An isolated baseline (`e799c2c`) versus candidate default `prepare-sources` fixture with N=2, `Release: 4%{?dist}`, and an existing `%changelog` produced byte-identical complete specs (SHA-256 `4a843e4fd3291c6a9f09585fe4189d074c5b575694595f5554f3095cd00d049a`; `cmp` exit 0). CLI references were generated by `mage docs`; no scenario snapshot update was needed.

## Deferred migration

The default intentionally remains legacy static calculation until Azure Linux migration is complete. The transactional runner and its `rpmdev-bumpspec`, `rpmdev-packager`, RPM, rpmspec, and Python RPM-module dependencies run only when the explicit preview flag is selected.

### DIFF
--- a/internal/app/azldev/core/sources/bumpspec.go
+++ b/internal/app/azldev/core/sources/bumpspec.go
@@ -1,0 +1,401 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package sources
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/microsoft/azure-linux-dev-tools/internal/global/opctx"
+	"github.com/microsoft/azure-linux-dev-tools/internal/projectconfig"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/externalcmd"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileperms"
+	"github.com/microsoft/azure-linux-dev-tools/internal/utils/fileutils"
+)
+
+const (
+	// RPMDevBumpspecBinary is the executable used to make one spec release bump.
+	RPMDevBumpspecBinary = "rpmdev-bumpspec"
+
+	rpmdevPackagerBinary = "rpmdev-packager"
+	rpmBinary            = "rpm"
+	rpmspecBinary        = "rpmspec"
+	python3Binary        = "python3"
+
+	rpmautospecNoBumpOutput = "RPMAutoSpec usage detected, not changing the spec file."
+
+	rpmdevBumpspecBinDirName = ".rpmdev-bumpspec-bin"
+	rpmdevBumpspecRPMWrapper = "rpm"
+
+	rpmdevBumpspecEVRSeparator   = "\x1f"
+	rpmdevBumpspecEVRQueryFormat = "%{NAME}\x1f%{EPOCHNUM}\x1f%{VERSION}\x1f%{RELEASE}\n"
+	rpmdevBumpspecEVRFieldCount  = 4
+	rpmLabelCompareProgram       = "import rpm, sys\n" +
+		"print(rpm.labelCompare(tuple(sys.argv[1:4]), tuple(sys.argv[4:7])))\n"
+)
+
+// ErrRPMDevBumpspecNoMutation is returned when rpmdev-bumpspec exits successfully without
+// producing a newer source package EVR.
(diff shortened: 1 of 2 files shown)


internal/app/azldev/core/sources/bumpspec.go
```diff
@@ -0,0 +1,401 @@
...
+	return "'" + strings.ReplaceAll(value, "'", "'\"'\"'") + "'"
+}
+
+func evaluateRPMDevBumpspecSourceEVR(ctx opctx.Ctx, request RPMDevBumpspecRequest) (RPMDevBumpspecSourceEVR, error) {
+	args := append(rpmdevBumpspecContextArgs(request),
+		"-q", "--srpm",
+		"--define", "dist %{nil}",
+		"--qf", rpmdevBumpspecEVRQueryFormat,
+		request.SpecPath,
+	)
```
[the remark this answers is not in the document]
trungams: that's quite unlikely to happen

internal/app/azldev/core/sources/bumpspec.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func shellQuote(value string) string {
+	return "'" + strings.ReplaceAll(value, "'", "'\"'\"'") + "'"
+}
+
+func evaluateRPMDevBumpspecSourceEVR(ctx opctx.Ctx, request RPMDevBumpspecRequest) (RPMDevBumpspecSourceEVR, error) {
+	args := append(rpmdevBumpspecContextArgs(request),
+		"-q", "--srpm",
+		"--define", "dist %{nil}",
```
liunan-ms, line 263: Why is `"--define", "dist %{nil}"` added here? With dist=nil here, the dist validation is disabled, would it be better to add the default dist only if dist is absent from the `rpmdevBumpspecContextArgs(request)`?

liunan-ms:
There're some conflicts to resolve, and also invite @Tonisal-byte to review the release bump changes.

internal/app/azldev/cmds/component/render.go
```diff
@@ -36,6 +36,7 @@ type RenderOptions struct {
 	Force             bool
 	CleanStale        bool
 	CheckOnly         bool
+	RPMDevBumpspec    bool
```
Tonisal-byte, line 39: I like the idea of this feature, and I believe it is the right way to increment static releases. However, I'm concerned we might hit a failure if we were to ever try to merge changes produced from the flag into the corerepo. I'm pretty sure there is some sort of spec drift check that would cause this to fail (since it runs without the `--rpmdev-bumpspec` flag.

Tonisal-byte: If we were to make rpmdev-bumpspec automatic, what sort of performance implications would it have?

trungams: > I like the idea of this feature, and I believe it is the right way to increment static releases. However, I'm concerned we might hit a failure if we were to ever try to merge changes produced from the flag into the corerepo. I'm pretty sure there is some sort of spec drift check that would cause this to fail (since it runs without the `--rpmdev-bumpspec` flag.

We definitely need follow-up PRs after this one to properly introduce the change to azurelinux repo, e.g we need to address changelog generation then update the spec drift check accordingly. I left those out of this PR to keep this initial change simple.

trungams: > If we were to make rpmdev-bumpspec automatic, what sort of performance implications would it have?

It's a standalone python script that doesn't require any special setup (e.g mock), so it should be fast enough and can be parallelized if needed

internal/app/azldev/core/sources/bumpspec.go
trungams: It's so the dist string won't have any effect on the EVR string. We are checking that the EVR strictly increases after bumpspec is invoked.

PR #356, opened 2026-09-17
feat(cli): add global concurrency control

ddstreet:
Add --concurrency as the shared base limit for parallel operations. The value defaults to the number of logical CPUs, with CPU-bound, I/O-bound, and fast worker pools using 1x, 2x, and 4x that value respectively.

Handle automatic, explicit, and clamped values consistently, add test coverage, and regenerate the CLI reference documentation.


ddstreet:
note that this is needed for many operations that involve remote actions (e.g. `refresh-upstream-commits`) where the system running `azldev` has a significant number of cpus. For example, on my 144-core system, when refreshing upstream commits for all components, azldev spawns 288 parallel threads hammering upstream fedora dist-git repos, which mostly all fail obviously due to upstream usage throttling.

trungams (approved):

trungams:
The change LGTM, but looks like some test snapshots need to be updated

liunan-ms (approved):
LGTM

